PR #12, opened 2017-05-09
Errors

Kangz:
PTAL, I'd like to get a first review before I start adding tests for this.


src/backend/common/Builder.h
```diff
@@ -0,0 +1,100 @@
...
+    //    called exactly once with an error, a success, or "unknown" if the
+    //    builder is destroyed; also the builder callback cannot be called before
+    //    either the object is destroyed or GetResult is called.
+    //
+    // It is possible for error to be generated before the error callback is
+    // registered when a builder "set" function performance validation inline.
+    // Because of this we have to store the status in the builder and defer
+    // calling the callback to GetResult.
+
+    class Builder : public RefCounted {
```
kainino0x: Seems like you're gonna have the same `GetResult` boilerplate everywhere; could defined in `Builder` instead if it were templated? `GetResultImpl` would be a virtual on `Builder`, `HandleResult` would use the template of the class.

Kangz: Addressed this, but in the patch that adds builder callbacks to the backens.

Kangz:
PTAL again, actually I think it would be better to land this before adding all the tests, so we can avoid having too much merge conflicts.

I added a couple additional commits that make sure no error goes silent for now.

Kangz:
Rebased, thanks for the reviews!

Issue #7, opened 2017-05-02
Introduce render passes


Kangz:
NXT uses a default framebuffer style of rendering as a placeholder for render passes. For automated testing we need to read-back what was rendered so we need to add render passes.

TODO (@kainino0x):
- [x] First design #14 
- [x] First implementation (Metal backend) #14 
- [x] Render-to-texture demo #14 
- [x] Update other demos #14 
- [x] Update glTFViewer #39 
- [ ] Check fragment shader output types matches render pass metadata
- [ ] Testing?
- [x] OpenGL backend #73 
- [ ] Any TODOs I added in #14 - especially the backbuffer hack (blocked on WSI)
- [ ] load/store operations

kainino0x:
First implementation is done in #14.

In particular:
* First API design (many features missing, including input attachments)
* Metal implementation (no OpenGL yet)
* Render-to-texture demo (a little broken until we have depth buffers)
* Update examples to use render passes

TODO:
- [ ] Check fragment output matches render pass
- [ ] Testing?
- [ ] OpenGL backend
- [ ] Any TODOs I added in #14 - especially the backbuffer hack (blocked on WSI)

PR #14, opened 2017-05-10
Render Passes

kainino0x:
There's a lot here. I think it's worth a first pass review finally. PTAL

EDIT: For TODOs, see Issue #7.

### DIFF
--- a/examples/Animometer.cpp
+++ b/examples/Animometer.cpp
@@ -22,6 +22,8 @@
 nxt::Device device;
 nxt::Queue queue;
 nxt::Pipeline pipeline;
+nxt::RenderPass renderpass;
+nxt::Framebuffer framebuffer;
 
 float RandomFloat(float min, float max) {
     float zeroOne = rand() / float(RAND_MAX);
@@ -103,7 +105,9 @@ void init() {
         })"
     );
 
+    CreateDefaultRenderPass(device, &renderpass, &framebuffer);
     pipeline = device.CreatePipelineBuilder()
+        .SetSubpass(renderpass, 0)
         .SetStage(nxt::ShaderStage::Vertex, vsModule, "main")
         .SetStage(nxt::ShaderStage::Fragment, fsModule, "main")
         .GetResult();
@@ -129,6 +133,7 @@ void frame() {
     for (int j = 0; j < 50; j++) {
 
         nxt::CommandBufferBuilder builder = device.CreateCommandBufferBuilder()
+            .BeginRenderPass(renderpass, framebuffer)
             .SetPipeline(pipeline)
             .Clone();
 
@@ -140,6 +145,7 @@ void frame() {
             i++;
         }
 
+        builder.EndRenderPass();
         commands[j] = builder.GetResult();
     }
 
--- a/examples/ComputeBoids.cpp
+++ b/examples/ComputeBoids.cpp
@@ -28,6 +28,8 @@ nxt::Buffer modelBuffer;
 std::array<nxt::Buffer, 2> particleBuffers;
 
 nxt::Pipeline renderPipeline;
+nxt::RenderPass renderpass;
+nxt::Framebuffer framebuffer;
 
 nxt::Buffer updateParams;
 nxt::Pipeline updatePipeline;
@@ -134,7 +136,9 @@ void initRender() {
         .SetInput(1, sizeof(glm::vec2), nxt::InputStepMode::Vertex)
         .GetResult();
 
+    CreateDefaultRenderPass(device, &renderpass, &framebuffer);
     renderPipeline = device.CreatePipelineBuilder()
+        .SetSubpass(renderpass, 0)
         .SetStage(nxt::ShaderStage::Vertex, vsModule, "main")
         .SetStage(nxt::ShaderStage::Fragment, fsModule, "main")
         .SetInputState(inputState)
@@ -279,11 +283,13 @@ void initCommandBuffers() {
             .SetBindGroup(0, updateBGs[i])
             .Dispatch(kNumParticles, 1, 1)
 
-            .SetPipeline(renderPipeline)
-            .TransitionBufferUsage(bufferDst, nxt::BufferUsageBit::Vertex)
-            .SetVertexBuffers(0, 1, &bufferDst, zeroOffsets)
-            .SetVertexBuffers(1, 1, &modelBuffer, zeroOffsets)
-            .DrawArrays(3, kNumParticles, 0, 0)
+            .BeginRenderPass(renderpass, framebuffer)
+                .SetPipeline(renderPipeline)
+                .TransitionBufferUsage(bufferDst, nxt::BufferUsageBit::Vertex)
+                .SetVertexBuffers(0, 1, &bufferDst, zeroOffsets)
+                .SetVertexBuffers(1, 1, &modelBuffer, zeroOffsets)
+                .DrawArrays(3, kNumParticles, 0, 0)
+            .EndRenderPass()
 
             .GetResult();
     }
--- a/examples/HelloCompute.cpp
+++ b/examples/HelloCompute.cpp
@@ -22,6 +22,8 @@ nxt::Queue queue;
 nxt::Buffer buffer;
 nxt::Pipeline renderPipeline;
 nxt::BindGroup renderBindGroup;
+nxt::RenderPass renderpass;
+nxt::Framebuffer framebuffer;
 nxt::Pipeline computePipeline;
 nxt::BindGroup computeBindGroup;
 
@@ -107,7 +109,9 @@ void init() {
             .SetBindGroupLayout(0, bgl)
             .GetResult();
 
+        CreateDefaultRenderPass(device, &renderpass, &framebuffer);
         renderPipeline = device.CreatePipelineBuilder()
+            .SetSubpass(renderpass, 0)
             .SetLayout(pl)
             .SetStage(nxt::ShaderStage::Vertex, vsModule, "main")
             .SetStage(nxt::ShaderStage::Fragment, fsModule, "main")
@@ -128,10 +132,12 @@ void frame() {
         .SetBindGroup(0, computeBindGroup)
         .Dispatch(1, 1, 1)
 
-        .SetPipeline(renderPipeline)
-        .TransitionBufferUsage(buffer, nxt::BufferUsageBit::Uniform)
-        .SetBindGroup(0, renderBindGroup)
-        .DrawArrays(3, 1, 0, 0)
+        .BeginRenderPass(renderpass, framebuffer)
+            .SetPipeline(renderPipeline)
+            .TransitionBufferUsage(buffer, nxt::BufferUsageBit::Uniform)
+            .SetBindGroup(0, renderBindGroup)
+            .DrawArrays(3, 1, 0, 0)
+        .EndRenderPass()
 
         .GetResult();
 
--- a/examples/HelloVertices.cpp
+++ b/examples/HelloVertices.cpp
@@ -23,6 +23,8 @@ nxt::Buffer vertexBuffer;
 
 nxt::Queue queue;
 nxt::Pipeline pipeline;
+nxt::RenderPass renderpass;
+nxt::Framebuffer framebuffer;
 
 void initBuffers() {
     static const float vertexData[12] = {
@@ -59,7 +61,7 @@ void init() {
 
     nxt::ShaderModule fsModule = CreateShaderModule(device, nxt::ShaderStage::Fragment, R"(
         #version 450
-        out vec4 fragColor;
+        layout(location = 0) out vec4 fragColor;
         void main() {
             fragColor = vec4(1.0, 0.0, 0.0, 1.0);
         })"
@@ -70,7 +72,9 @@ void init() {
         .SetInput(0, 4 * sizeof(float), nxt::InputStepMode::Vertex)
         .GetResult();
 
+    CreateDefaultRenderPass(device, &renderpass, &framebuffer);
     pipeline = device.CreatePipelineBuilder()
+        .SetSubpass(renderpass, 0)
         .SetStage(nxt::ShaderStage::Vertex, vsModule, "main")
         .SetStage(nxt::ShaderStage::Fragment, fsModule, "main")
         .SetInputState(inputState)
@@ -80,9 +84,11 @@ void init() {
 void frame() {
     static const uint32_t vertexBufferOffsets[1] = {0};
     nxt::CommandBuffer commands = device.CreateCommandBufferBuilder()
-        .SetPipeline(pipeline)
-        .SetVertexBuffers(0, 1, &vertexBuffer, vertexBufferOffsets)
-        .DrawArrays(3, 1, 0, 0)
+        .BeginRenderPass(renderpass, framebuffer)
+            .SetPipeline(pipeline)
+            .SetVertexBuffers(0, 1, &vertexBuffer, vertexBufferOffsets)
+            .DrawArrays(3, 1, 0, 0)
+        .EndRenderPass()
         .GetResult();
 
     queue.Submit(1, &commands);
--- a/next.json
+++ b/next.json
@@ -219,6 +219,19 @@
                 "returns": "command buffer"
             },
             {
+                "name": "begin render pass",
+                "args": [
+                    {"name": "render pass", "type": "render pass"},
+                    {"name": "framebuffer", "type": "framebuffer"}
+                ]
+            },
+            {
+                "name": "advance subpass"
+            },
+            {
+                "name": "end render pass"
+            },
+            {
                 "name": "copy buffer to texture",
                 "args": [
                     {"name": "buffer", "type": "buffer"},
@@ -348,6 +361,10 @@
                 "returns": "command buffer builder"
             },
             {
+                "name": "create framebuffer builder",
+                "returns": "framebuffer builder"
+            },
+            {
                 "name": "create input state builder",
                 "returns": "input state builder"
             },
@@ -362,6 +379,10 @@
             {
                 "name": "create queue builder",
                 "returns": "queue builder"
+            },
+            {
+                "name": "create render pass builder",
+                "returns": "render pass builder"
             },
             {
                 "name": "create sampler builder",
@@ -401,6 +422,38 @@
         "values": [
             {"value": 0, "name":"nearest"},
             {"value": 1, "name":"linear"}
+        ]
+    },
+    "framebuffer": {
+        "category": "object"
+    },
+    "framebuffer builder": {
+        "category": "object",
+        "methods": [
+            {
+                "name": "get result",
+                "returns": "framebuffer"
+            },
+            {
+                "name": "set render pass",
+                "args": [
+                    {"name": "render pass", "type": "render pass"}
+                ]
+            },
+            {
+                "name": "set dimensions",
+                "args": [
+                    {"name": "width", "type": "uint32_t"},
+                    {"name": "height", "type": "uint32_t"}
+                ]
+            },
+            {
+                "name": "set attachment",
+                "args": [
+                    {"name": "attachment slot", "type": "uint32_t"},
+                    {"name": "texture views", "type": "texture view"}
+                ]
+            }
         ]
     },
     "index format": {
@@ -463,6 +516,13 @@
                 ]
             },
             {
+                "name": "set subpass",
+                "args": [
+                    {"name": "render pass", "type": "render pass"},
+                    {"name": "subpass", "type": "uint32_t"}
+                ]
+            },
+            {
                 "name": "set stage",
                 "args": [
                     {"name": "stage", "type": "shader stage"},
@@ -517,6 +577,50 @@
                 "returns": "queue"
             }
         ]
+    },
+    "render pass builder": {
+        "category": "object",
+        "TODO": {
+            "attachments": "Also need load op, store op, depth/stencil attachments and depth/stencil ops, and maybe usages for the implicit attachment transitions",
+            "subpasses": "Also need input attachments, resolve attachments, and preserve attachments"
+        },
+        "methods": [
+            {
+                "name": "get result",
+                "returns": "render pass"
+            },
+            {
+                "name": "set attachment count",
+                "args": [
+                    {"name": "attachment count", "type": "uint32_t"}
+                ]
+            },
+            {
+                "name": "attachment set format",
+                "TODO": "Also need sample count",
+                "args": [
+                    {"name": "attachment slot", "type": "uint32_t"},
+                    {"name": "format", "type": "texture format"}
+                ]
+            },
+            {
+                "name": "set subpass count",
+                "args": [
+                    {"name": "subpass count", "type": "uint32_t"}
+                ]
+            },
+            {
+                "name": "subpass set color attachment",
+                "args": [
+                    {"name": "subpass index", "type": "uint32_t"},
+                    {"name": "output attachment location", "type": "uint32_t"},
+                    {"name": "attachment slot", "type": "uint32_t"}
+                ]
+            }
+        ]
+    },
+    "render pass": {
+        "category": "object"
     },
     "sampler": {
         "category": "object"
--- a/src/backend/common/CommandBuffer.cpp
+++ b/src/backend/common/CommandBuffer.cpp
@@ -55,6 +55,18 @@ namespace backend {
         Command type;
         while(commands->NextCommandId(&type)) {
             switch (type) {
+                case Command::AdvanceSubpass:
+                    {
+                        AdvanceSubpassCmd* cmd = commands->NextCommand<AdvanceSubpassCmd>();
+                        cmd->~AdvanceSubpassCmd();
+                    }
+                    break;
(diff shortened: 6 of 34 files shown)


Kangz:
FYI I merged the errors PR so you'll need to do a couple changes:
 - Builders inherit from Builder<ObjectBase>
 - Builders GetResult becomes GetResultImpl() overrride and is private
 - Most cases of device->HandleError become HandleError

next.json
```diff
@@ -493,6 +547,42 @@
...
+            {
+                "name": "enable attachment",
+                "comment": "Lazy implementation; probably should make an Attachment object.",
+                "args": [
+                    {"name": "attachment slot", "type": "uint32_t"},
+                    {"name": "format", "type": "texture format"}
+                ]
+            },
+            {
+                "name": "set subpass count",
```
Kangz: Likewise we should probably have a "set attachment count" and validate that both are done before anything else?

src/backend/common/CommandBuffer.cpp
```diff
@@ -238,9 +257,75 @@ namespace backend {
...
+                case Command::BeginRenderPass:
+                    {
+                        BeginRenderPassCmd* begin = iterator.NextCommand<BeginRenderPassCmd>();
+                        auto* renderPass = begin->renderPass.Get();
+                        auto* framebuffer = begin->framebuffer.Get();
+                        if (currentRenderPass != nullptr) {
+                            device->HandleError("A render pass is already active");
+                            return false;
+                        }
+                        if (framebuffer->GetRenderPass() != renderPass) {
```
Kangz: Leave a TODO that this compatibility check will need to be slightly relaxed at some point, to allow render passes with different load and store ops.

src/backend/common/CommandBuffer.cpp
```diff
@@ -238,9 +257,75 @@ namespace backend {
...
+                            device->HandleError("Framebuffer is incompatible with this render pass");
+                            return false;
+                        }
+                        lastPipeline = nullptr;
```
Kangz: What the reason to reset the pipeline and bindgroups? I think there could be a use for inheriting bindgroups from outside the pipeline.

I'm not sure what the best way to go is, but this is inconsistent with endrenderpass that also reset vertex and index buffers.

Kangz: Also we could add a validation aspect for renderpasses.

src/backend/common/CommandBuffer.cpp
```diff
@@ -358,21 +443,44 @@ namespace backend {
                     }
                     break;
 
+                case Command::EndRenderPass:
+                    {
+                        iterator.NextCommand<EndRenderPassCmd>();
+                        if (currentRenderPass == nullptr) {
+                            device->HandleError("No render pass is currently active");
+                            return false;
+                        }
```
Kangz: We should also check that we are at the last subpass.

src/backend/common/CommandBuffer.cpp
```diff
@@ -358,21 +443,44 @@ namespace backend {
...
-                            aspects.reset(VALIDATION_ASPECT_RENDER_PIPELINE);
                         } else {
+                            if (currentRenderPass == nullptr) {
+                                device->HandleError("A render pass must be active when a render pipeline is set");
+                                return false;
+                            }
                             aspects.set(VALIDATION_ASPECT_RENDER_PIPELINE);
-                            aspects.reset(VALIDATION_ASPECT_COMPUTE_PIPELINE);
                         }
-                        aspects.reset(VALIDATION_ASPECT_BINDGROUPS);
```
Kangz: This dirties the validation of vertex buffers and bindgroups so that we can perform it again at draw time. It should be kept.

src/backend/common/CommandBuffer.cpp
```diff
@@ -476,6 +584,15 @@ namespace backend {
                             return false;
                         }
 
+                        auto& subpassInfo = currentRenderPass->GetSubpassInfo(currentSubpass);
```
Kangz: In the object introduced to handle all the usage tracking, this could be done with a "temporarily frozen" tag put on resources.

src/backend/common/Forward.h
```diff
@@ -64,6 +68,9 @@ namespace backend {
     static constexpr uint32_t kMaxVertexAttributes = 16u;
     static constexpr uint32_t kMaxVertexInputs = 16u;
     static constexpr uint32_t kNumStages = 3;
+    static constexpr uint32_t kMaxTextureViewsPerFramebuffer = 16u;
```
Kangz: Is there a practical reason for this, like a Vulkan limitation? Otherwise Framebuffer could require the renderpass to be given before the texture views, and resize a vector.

src/backend/common/Forward.h
```diff
@@ -64,6 +68,9 @@ namespace backend {
...
+    static constexpr uint32_t kMaxRenderPassAttachments = 16u;
```
Kangz: ditto, but for renderpasses

src/backend/common/Framebuffer.cpp
```diff
@@ -0,0 +1,102 @@
...
+    bool FramebufferBuilder::WasConsumed() const {
+        return consumed;
+    }
+
+    FramebufferBase* FramebufferBuilder::GetResult() {
+        constexpr int allProperties = FRAMEBUFFER_PROPERTY_RENDER_PASS | FRAMEBUFFER_PROPERTY_DIMENSIONS;
+        if ((propertiesSet & allProperties) != allProperties) {
+            device->HandleError("Framebuffer missing properties");
+            return nullptr;
+        }
```
Kangz: Also needs to check all attachments required by the renderpass are set, that is, that no element of the vector is null.

src/backend/common/Framebuffer.h
```diff
@@ -0,0 +1,74 @@
...
+        private:
+            friend class FramebufferBase;
+
+            DeviceBase* device = nullptr;
+            Ref<RenderPassBase> renderPass;
+            uint32_t width = 0;
+            uint32_t height = 0;
+            std::array<Ref<TextureViewBase>, kMaxTextureViewsPerFramebuffer> textureViews;
+            int propertiesSet = 0;
+            bool consumed = false;
```
Kangz: Also with the merge of the errors PR, all the "consumed" disappeared

src/backend/common/RenderPass.h
```diff
@@ -0,0 +1,72 @@
...
+#include <vector>
+
+namespace backend {
+
+    class RenderPassBase : public RefCounted {
+        public:
+            RenderPassBase(RenderPassBuilder* builder);
+
+            struct SubpassInfo {
+                std::map<uint32_t, uint32_t> colorAttachments;
```
Kangz: std::array<kMaxColorAttachments, uint32_t>
will need to update the code in CommandBuffer that iterates over it to check for nulltpr

src/backend/common/RenderPass.cpp
```diff
@@ -0,0 +1,103 @@
...
+    void RenderPassBuilder::SetSubpassColorAttachment(uint32_t subpassIndex, uint32_t outputAttachmentLocation, uint32_t attachmentSlot) {
+        if (subpassIndex >= subpasses.size()) {
+            device->HandleError("Subpass index out of bounds");
+            return;
+        }
+        if (outputAttachmentLocation >= kMaxColorAttachments) {
+            device->HandleError("Output attachment location out of bounds");
+            return;
+        }
+        if (!attachmentsEnabled[attachmentSlot]) {
```
Kangz: I think we can force attachments to be a full and contiguous range so this check disappears.

src/backend/common/RenderPass.cpp
```diff
@@ -0,0 +1,103 @@
...
+            device->HandleError("Attachment slot not enabled");
+            return;
+        }
+
```
Kangz: Also check we are not setting it a second time.

src/backend/common/RenderPass.cpp
```diff
@@ -0,0 +1,103 @@
...
+    RenderPassBuilder::RenderPassBuilder(DeviceBase* device)
+        : device(device), subpasses(1) {
+    }
+
+    bool RenderPassBuilder::WasConsumed() const {
+        return consumed;
+    }
+
+    RenderPassBase* RenderPassBuilder::GetResult() {
+        consumed = true;
```
Kangz: We'll need to check a whole lot of things here once we add preserved attachments, and stuff like that.

But also for now we should check that all the attachments and subpasses have been defined (I don't see a usecase for sparse index spaces).

src/backend/common/CommandBuffer.cpp
```diff
@@ -238,9 +257,75 @@ namespace backend {
...
+                        currentFramebuffer = framebuffer;
+                        currentSubpass = 0;
+                        if (!beginSubpass()) {
+                            return false;
+                        }
+                    }
+                    break;
+
                 case Command::CopyBufferToTexture:
                     {
```
Kangz: Blits should be forbidden inside renderpasses

src/backend/common/Framebuffer.cpp
Kangz: Or count that there is at most one unset attachment for the "default framebuffer" hack and add a TODO ;)

src/backend/common/RenderPass.h
kainino0x: Had to also add a `std::bitset<kMaxColorAttachments> colorAttachmentsSet` - I think this is reasonable.

src/backend/common/RenderPass.cpp
kainino0x: This is sort of a fuzzy concept as these could have optional properties with default values, but I've added this validation for the format field of attachments at least. A subpass by default will just not have any attachments.

next.json
```diff
@@ -493,6 +547,42 @@
...
+                "comment": "Lazy implementation; probably should make a Subpass object.",
+                "args": [
+                    {"name": "subpass count", "type": "uint32_t"}
+                ]
+            },
+            {
+                "name": "set subpass color attachment",
+                "comment": "Lazy implementation; probably should make a Subpass object.",
```
Kangz: You mean as in not having a dedicated subpass object? Sounds good.

src/backend/common/CommandBuffer.cpp
```diff
@@ -238,9 +257,75 @@ namespace backend {
...
         };
 
+        // TODO(kainino@chromium.org): Manage this state inside an object, change lambda to a method
+        RenderPassBase* currentRenderPass = nullptr;
+        FramebufferBase* currentFramebuffer = nullptr;
+        uint32_t currentSubpass = 0;
+        auto beginSubpass = [&]() -> bool {
+            auto& subpassInfo = currentRenderPass->GetSubpassInfo(currentSubpass);
+            for (auto& kv : subpassInfo.colorAttachments) {
+                if (auto* tv = currentFramebuffer->GetTextureView(kv.second)) {
```
Kangz: Yep realized that afterwards, but forgot to edit this comment, TODO sounds good.

src/backend/common/RenderPass.h
Kangz: Bitset sounds good.

examples/HelloVertices.cpp
```diff
@@ -80,8 +80,9 @@ void init() {
 void frame() {
     static const uint32_t vertexBufferOffsets[1] = {0};
     nxt::RenderPass renderpass = device.CreateRenderPassBuilder()
-        .EnableAttachment(0, nxt::TextureFormat::R8G8B8A8Unorm)
-        .SetSubpassColorAttachment(0, 0, 0)
+        .SetAttachmentCount(1)
+        .AttachmentSetFormat(0, nxt::TextureFormat::R8G8B8A8Unorm)
+        .SubpassSetColorAttachment(0, 0, 0)
```
Kangz: There is a missing SetSubpassCount here, we should double check validation catches this case.

next.json
```diff
@@ -423,11 +423,16 @@
                 ]
             },
             {
-                "name": "set attachments",
+                "name": "set attachment count",
```
Kangz: The attachment count can implicitly be the same one as the renderpass.

src/backend/common/Framebuffer.cpp
```diff
@@ -52,12 +53,28 @@ namespace backend {
...
+                if (textureView == nullptr) {
+                    if (usingBackbufferHack) {
+                        device->HandleError("Some subpass uses more than one texture view which is null in this framebuffer");
+                        return nullptr;
+                    }
+                    usingBackbufferHack = true;
+                }
+            }
+        }
+
```
Kangz: nit: check format and samples match the renderpass.

src/backend/common/CommandBuffer.cpp
```diff
@@ -597,14 +598,25 @@ namespace backend {
...
 
-                        // TODO(kainino@chromium.org): We should be able to optimize this by tracking which textures are in use as attachments.
-                        auto& subpassInfo = currentRenderPass->GetSubpassInfo(currentSubpass);
-                        for (auto attachmentSlot : subpassInfo.colorAttachments) {
-                            // XXX: this loop is running too long; how do we make it work correctly?
-                            auto* tv = currentFramebuffer->GetTextureView(attachmentSlot);
-                            if (tv && tv->GetTexture() == texture) {
-                                device->HandleError("Can't transition a texture while it's used as a color attachment");
-                                return false;
+                        // TODO(kainino@chromium.org): We should be able to
```
Kangz: I'm thinking that we could have scratch buffers in textures and buffers, that can be used during validation to store custom data such as this. Validation looks good like this for now.

src/backend/common/RenderPass.cpp
Kangz: Yeah this is a bit fuzzy, I'll think about this more.

src/backend/common/RenderPass.cpp
Kangz: I don't see the bitset being checked.

Kangz:
Added a couple comments but reviewing the incremental changes is a bit hard, I'll do a full re-review too.

examples/HelloVertices.cpp
kainino0x: Currently, subpass count defaults to 1, attachment count defaults to 0. They aren't required.

next.json
kainino0x: Thanks, good point.

src/backend/common/Framebuffer.cpp
kainino0x: Added this validation in SetAttachment

src/backend/common/RenderPass.cpp
kainino0x: Done

src/backend/common/RenderPass.cpp
kainino0x: It's there for me, this diff is outdated (ergh github...)

kainino0x:
Sorry, I take that back, there's still one more thing I need to do (that I forgot about after the weekend).

kainino0x:
Okay, fixed some bugs and update the examples. Now it's ready.

BTW I'm having some issues with error handling - most API usage errors (HandleError) seem to cause segfaults due to dereferencing null built objects. Maybe there's an issue in my code, maybe in the error handling code, not sure.

src/backend/common/Pipeline.cpp
```diff
@@ -33,6 +35,12 @@ namespace backend {
             return;
         }
 
+        if (!IsCompute() && !renderPass) {
+            builder->HandleError("Pipeline render pass not set");
+            return;
+        }
+        // TODO(kainino@chromium.org): Need to verify the pipeline against its render subpass.
```
kainino0x: This is a bit of work because the info is (afaik) only available in ExtractSpirvInfo, so I'm going to make it a separate patch.

src/backend/common/CommandBuffer.cpp
```diff
@@ -472,6 +594,28 @@ namespace backend {
                             return false;
                         }
 
+                        // TODO(kainino@chromium.org): We should be able to
+                        // optimize this by tracking which textures are in use
+                        // as attachments. Maybe it's possible that the
+                        // XAttachment usages could mark this: disallow
```
Kangz: note: This could work and this would mean the end usage will have to be encoded in the renderpass. I'm not 100% sure this is desirable but don't see a counter example.

src/backend/common/CommandBuffer.cpp
```diff
@@ -472,6 +594,28 @@ namespace backend {
...
+                        // in an XAttachment usage.
+                        if (currentRenderPass) {
+                            auto& info = currentRenderPass->GetSubpassInfo(currentSubpass);
+                            for (uint32_t location = 0; location < info.colorAttachments.size(); ++location) {
+                                if (!info.colorAttachmentsSet[location]) {
+                                    continue;
+                                }
+                                uint32_t attachmentSlot = info.colorAttachments[location];
+                                auto* tv = currentFramebuffer->GetTextureView(attachmentSlot);
+                                if (tv && tv->GetTexture() == texture) {
```
Kangz: nit: TODO, check for tv != nullptr will not be needed once we have WSI

src/backend/common/Framebuffer.cpp
```diff
@@ -0,0 +1,125 @@
...
+            return;
+        }
+
+        this->width = width;
+        this->height = height;
+        propertiesSet |= FRAMEBUFFER_PROPERTY_DIMENSIONS;
+    }
+
+    void FramebufferBuilder::SetAttachment(uint32_t attachmentSlot, TextureViewBase* textureView) {
+        if ((propertiesSet & FRAMEBUFFER_PROPERTY_RENDER_PASS) != FRAMEBUFFER_PROPERTY_RENDER_PASS) {
```
Kangz: nit: properties & RENDER_PASS == 0

src/backend/common/Framebuffer.cpp
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    FramebufferBase* FramebufferBuilder::GetResultImpl() {
+        constexpr int requiredProperties = FRAMEBUFFER_PROPERTY_RENDER_PASS | FRAMEBUFFER_PROPERTY_DIMENSIONS;
+        if ((propertiesSet & requiredProperties) != requiredProperties) {
+            HandleError("Framebuffer missing properties");
+            return nullptr;
+        }
+
+        for (uint32_t subpass = 0; subpass < renderPass->GetSubpassCount(); ++subpass) {
```
Kangz: Since the renderpass's attachments must all be defined, we only need to iterate over this->textureViews.

src/backend/common/Pipeline.cpp
Kangz: Souds good

src/backend/common/CommandBuffer.cpp
```diff
@@ -354,18 +452,42 @@ namespace backend {
...
+                                return false;
+                            }
                             aspects.set(VALIDATION_ASPECT_COMPUTE_PIPELINE);
-                            aspects.reset(VALIDATION_ASPECT_RENDER_PIPELINE);
                         } else {
+                            if (!currentRenderPass) {
+                                device->HandleError("A render pass must be active when a render pipeline is set");
+                                return false;
+                            }
                             aspects.set(VALIDATION_ASPECT_RENDER_PIPELINE);
```
Kangz: nit: need to check that the pipeline's renderpass is compatible with currentRenderpass.

src/backend/common/RenderPass.cpp
```diff
@@ -0,0 +1,138 @@
...
+    enum RenderPassSetProperties {
+        RENDERPASS_PROPERTY_ATTACHMENT_COUNT = 0x1,
+        RENDERPASS_PROPERTY_SUBPASS_COUNT = 0x2,
+    };
+
+    RenderPassBuilder::RenderPassBuilder(DeviceBase* device)
+        : Builder(device), subpasses(1) {
+    }
+
+    RenderPassBase* RenderPassBuilder::GetResultImpl() {
```
Kangz: check that both renderpass count and attachment count are set. This will trigger errors in some of the samples.

Kangz: check that both the subpass count and the attachment counts are set. Some samples will need to be fixed.

src/backend/common/RenderPass.h
```diff
@@ -0,0 +1,87 @@
...
+        private:
+            friend class RenderPassBase;
+
+            enum AttachmentProperty {
+                ATTACHMENT_PROPERTY_FORMAT,
+                ATTACHMENT_PROPERTY_COUNT
+            };
+
+            enum SubpassProperty {
+                SUBPASS_PROPERTY_COUNT
```
Kangz: Do you expect many properties for subpasses and attachments? I'm wondering if just a std::vector<bool> would be enough.

examples/Animometer.cpp
```diff
@@ -22,6 +22,8 @@
 nxt::Device device;
 nxt::Queue queue;
 nxt::Pipeline pipeline;
+nxt::RenderPass renderpass;
```
Kangz: nit: maybe these "default" renderpass and FBO could be put in Utils

examples/Animometer.cpp
```diff
@@ -103,7 +105,17 @@ void init() {
         })"
     );
 
+    renderpass = device.CreateRenderPassBuilder()
+        .SetAttachmentCount(1)
+        .AttachmentSetFormat(0, nxt::TextureFormat::R8G8B8A8Unorm)
```
Kangz: nit: needs subpass count

examples/ComputeBoids.cpp
```diff
@@ -134,7 +136,17 @@ void initRender() {
         .SetInput(1, sizeof(glm::vec2), nxt::InputStepMode::Vertex)
         .GetResult();
 
+    renderpass = device.CreateRenderPassBuilder()
+        .SetAttachmentCount(1)
+        .AttachmentSetFormat(0, nxt::TextureFormat::R8G8B8A8Unorm)
+        .SubpassSetColorAttachment(0, 0, 0)
```
Kangz: ditto

examples/HelloCompute.cpp
```diff
@@ -107,7 +109,17 @@ void init() {
             .SetBindGroupLayout(0, bgl)
             .GetResult();
 
+        renderpass = device.CreateRenderPassBuilder()
+            .SetAttachmentCount(1)
+            .AttachmentSetFormat(0, nxt::TextureFormat::R8G8B8A8Unorm)
```
Kangz: ditto, and etc in other samples

src/backend/common/RenderPass.cpp
kainino0x: TBD

src/backend/common/RenderPass.h
Kangz: I meant ```std::vector<bool>``` for some reason GitHub ate the template param. It sounds that most subpass things could have obvious defaults, but not attachments, so AttachmentProperty looks needed.

Thanks!

kainino0x: Oh, I see re:`<bool>`.

I don't think defaults for subpass attachments make that much sense actually, but if we want to track them it's going to be more complex than a vector of bitsets, so I've removed SubpassProperty.

examples/ComputeBoids.cpp
kainino0x: Done

examples/HelloCompute.cpp
kainino0x: Done

src/backend/common/RenderPass.cpp
kainino0x: Done

examples/Animometer.cpp
kainino0x: Good idea, I'll do this

kainino0x: I put just the setup in Utils `CreateDefaultRenderPass` to initialize a render pass and framebuffer.

kainino0x:
Addressed comments and also squashed everything since I have no idea how well GitHub handles squashing merge commits.

src/backend/common/CommandBuffer.cpp
kainino0x: I was thinking of disallowing explicit transitions INTO XAttachment, but having render passes always leave their attachments in the 'no usage' state (if they're not frozen). Then we don't have to set the end usage in the render pass.

kainino0x: (Or leaving their attachments in the 'XAttachment' usage, because it doesn't matter whether it's that or 'no usage', I think.)

Kangz:
Just to be sure, do you want to merge this now or would you rather wait until the OpenGL backend is done? (and potentially other checkboxes)

kainino0x:
Hm, I think it would be better to merge now. I don't want to let it diverge too much, and it's going to take me a while to do the remaining tasks (been focusing on Chrome for the past few days).

Kangz:
LGTM, an ultra-nit is that "Squashed commit of ..." isn't the nicest. Maybe it could be "introduce renderpasses" instead?

kainino0x:
Rewrote commit message

Kangz:
LGTM

PR #16, opened 2017-05-17
First batch of new wire tests.

Kangz:
PTAL, this also makes it so all the unittests are in the same binary.

I'll keep making wire tests until the TODO at the end of WireTests.cpp is finished.


Kangz:
PTAL, finished adding all the wire tests I wanted to add

src/tests/unittests/WireTests.cpp
```diff
@@ -0,0 +1,493 @@
...
+        .WillOnce(Return(apiBuilder));
+
+    EXPECT_CALL(api, SamplerBuilderSetFilterMode(apiBuilder, NXT_FILTER_MODE_LINEAR, NXT_FILTER_MODE_LINEAR, NXT_FILTER_MODE_NEAREST))
+        .Times(1);
+
+    FlushClient();
+}
+
+// Test that the wire is able to send arrays of numerical values
+static uint32_t testPushConstantValues[4] = {
```
kainino0x: make this const or constexpr?

src/tests/unittests/WireTests.cpp
```diff
@@ -0,0 +1,493 @@
...
+
+    FlushClient();
+
+    EXPECT_CALL(*mockBuilderErrorCallback, Call(NXT_BUILDER_ERROR_STATUS_ERROR, _, 1, 2)).Times(1);
+    EXPECT_CALL(*mockBuilderErrorCallback, Call(NXT_BUILDER_ERROR_STATUS_ERROR, _, 3, 4)).Times(1);
+
+    FlushServer();
+}
+
+// Test that we get a succes builder error status when no error happens
```
kainino0x: "succes"

Kangz:
Thanks for the review!

PR #17, opened 2017-05-23
Factor out resource usage tracking into a class

kainino0x:
I'm planning to reuse this class for tracking usage state within command buffer builds.

### DIFF
--- a/src/backend/common/UsageTracker.h
+++ b/src/backend/common/UsageTracker.h
@@ -1,0 +1,81 @@
+// Copyright 2017 The NXT Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BACKEND_COMMON_USAGETRACKER_H
+#define BACKEND_COMMON_USAGETRACKER_H
+
+template<typename T, typename B>
+class UsageTracker {
+    public:
+        UsageTracker(B allowedUsage, B usage)
+            : allowedUsage(allowedUsage), currentUsage(usage) {
+        }
+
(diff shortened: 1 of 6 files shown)


src/backend/common/UsageTracker.h
```diff
@@ -0,0 +1,81 @@
...
+            }
+            return T::IsUsagePossible(allowedUsage, usage);
+        }
+
+        void TransitionUsageImpl(B usage) {
+            assert(IsTransitionPossible(usage));
+            currentUsage = usage;
+        }
+
+        bool ClearUsage() {
```
Kangz: nit: ClearUsage isn't used here or in the two other PRs, what is the expected use of it?

src/backend/common/UsageTracker.h
```diff
@@ -0,0 +1,81 @@
+// Copyright 2017 The NXT Authors
```
Kangz: Currently the usage tracking is the same for textures and buffers and this factorization makes sense, but textures will eventually have per-mip tracking, so this class will have to be split again. When that happens I think it will make sense to fold the UsageTracker in the resource class again.

kainino0x: That's a good point I forgot about. In that case I think I want to have two usage tracker classes. I want to be able to use the usage tracker classes by themselves (without resources) in the command buffer state tracker (for tracking usages locally).

src/backend/common/UsageTracker.h
kainino0x: Ah, this isn't used anymore. I was using it on endSubpass but realized that needed to be tracked locally in `mostRecentTextureUsages` instead.

src/backend/common/UsageTracker.h
Kangz: I'd be interested to see how tracking inside the command buffer and outside can be factored, it seems to me the additional tracking inside is much simpler (no freezing involved).
Also I'm not a huge fan of using maps to store temporary information. What do you think of having a scratch buffer inside resources that can be used to store information temporarily for validation? This way we would avoid allocations, hashing pointers and maybe reduce the number of cache lines we access.

kainino0x: That's a good point, it is a bit simpler. Right now it has its own implementation in the state tracking which isn't that long (though it keeps getting longer). Merging it into the resource objects was another approach I thought of. That's fine as long as we're careful not to leak transient state between command buffer validations.

kainino0x:
Decided offline not to merge this.

PR #18, opened 2017-05-23
Forbid explicit transitions to/from attachment usages

kainino0x:


src/backend/common/CommandBuffer.cpp
```diff
@@ -260,7 +273,11 @@ namespace backend {
         uint32_t currentSubpass = 0;
         auto beginSubpass = [&]() -> bool {
             auto& subpassInfo = currentRenderPass->GetSubpassInfo(currentSubpass);
-            for (auto attachmentSlot : subpassInfo.colorAttachments) {
+            for (size_t location = 0; location < subpassInfo.colorAttachments.size(); ++location) {
```
Kangz: nit: for(auto location: IterateBitSet(subpassInfo.colorAttachmentSet))

src/backend/common/CommandBuffer.cpp
```diff
@@ -278,11 +295,34 @@ namespace backend {
...
                 }
-                mostRecentTextureUsages.erase(texture);
+                mostRecentTextureUsages[texture] = nxt::TextureUsageBit::ColorAttachment;
                 texturesTransitioned.insert(texture);
             }
             return true;
         };
+        auto endSubpass = [&]() {
+            auto& subpassInfo = currentRenderPass->GetSubpassInfo(currentSubpass);
+            for (size_t location = 0; location < subpassInfo.colorAttachments.size(); ++location) {
```
Kangz: nit: use IterateBitSet here too

src/backend/common/CommandBuffer.cpp
```diff
@@ -201,6 +201,19 @@ namespace backend {
             auto it = mostRecentTextureUsages.find(texture);
             return it != mostRecentTextureUsages.end() && (it->second & usage);
         };
+        auto isTextureTransitionPossible = [&](TextureBase* texture, nxt::TextureUsageBit usage) -> bool {
+            const nxt::TextureUsageBit attachmentUsages =
+                nxt::TextureUsageBit::ColorAttachment |
+                nxt::TextureUsageBit::DepthStencilAttachment;
+            ASSERT(usage != nxt::TextureUsageBit::None && nxt::HasZeroOrOneBits(usage));
+            auto it = mostRecentTextureUsages.find(texture);
```
Kangz: This checks makes it so the following produces an error, inside our outside of a renderpass:

```cpp
cmdBufBuilder->TransitionTextureUsage(tex, nxt::TextureUsageBit::ColorAttachment);
cmdBufBuilder->TransitionTextureUsage(tex, nxt::TextureUsageBit::Sampled);
```

This is surprising behavior, I think it would be simpler to specify that the texture is "temporarily" frozen when it is a (color / DS / input) attachment for the current subpass.

kainino0x: Oh, I forgot to fix this case. I was going to also forbid explicit transitions INTO attachment usages. WDYT?

Kangz: Looks better with the transition into attachments forbidden. We are starting to build a bunch of special cases, if we need to write a spec for NXT I think we'll need to revisit the transitions rule a bit to see if they can be simplified. That said it is fine for now.

kainino0x:
Merging this for now but we're probably going to change the semantics here to make all transitions explicit.

PR #19, opened 2017-05-23
Create a command buffer builder state tracker object

kainino0x:
Dependent on #18, so this will be easier to review after that's merged and this is rebased. Defer if you want.

Later (after this lands) I plan to move more of the logic from CommandBuffer.cpp to CommandBufferStateTracker.cpp.

### DIFF
--- a/src/backend/common/CommandBufferStateTracker.cpp
+++ b/src/backend/common/CommandBufferStateTracker.cpp
@@ -1,0 +1,500 @@
+// Copyright 2017 The NXT Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "CommandBufferStateTracker.h"
+
+#include "Forward.h"
+#include "BindGroup.h"
+#include "BindGroupLayout.h"
+#include "BitSetIterator.h"
+#include "Buffer.h"
+#include "Framebuffer.h"
+#include "InputState.h"
+#include "Pipeline.h"
+#include "PipelineLayout.h"
+#include "RenderPass.h"
+#include "Texture.h"
+
+namespace backend {
+    CommandBufferStateTracker::CommandBufferStateTracker(CommandBufferBuilder* builder)
+        : builder(builder) {
+    }
+
+    bool CommandBufferStateTracker::HaveRenderPass() const {
+        return currentRenderPass != nullptr;
+    }
+
+    bool CommandBufferStateTracker::ValidateCanCopy() const {
+        if (currentRenderPass) {
+            builder->HandleError("Copy cannot occur during a render pass");
+            return false;
+        }
+        return true;
+    }
+
+    bool CommandBufferStateTracker::ValidateCanUseBufferAs(BufferBase* buffer, nxt::BufferUsageBit usage) const {
+        if (!BufferHasGuaranteedUsageBit(buffer, usage)) {
+            builder->HandleError("Buffer is not in the necessary usage");
+            return false;
+        }
+        return true;
+    }
+
+    bool CommandBufferStateTracker::ValidateCanUseTextureAs(TextureBase* texture, nxt::TextureUsageBit usage) const {
+        if (!TextureHasGuaranteedUsageBit(texture, usage)) {
+            builder->HandleError("Texture is not in the necessary usage");
+            return false;
+        }
+        return true;
+    }
+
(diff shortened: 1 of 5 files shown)


Kangz:
Overall I feel the separation of concerns between the tracker and CommandBufferBuilder isn't clear enough. Ideally all the persistent state would go in the tracker (include ```aspects```) and would be  private to the tracker.

Then code in the CommandBufferBuilder could look like these examples:

```cpp
// For stuff changing the state
if (!state->BeginRenderPass(renderpass, framebuffer)) {
  return false;
}
// stuff that doesn't change the state would look like this, example for blits
if (!state->ValidateCanCopy() ||
    !state->ValidateUseBufferAs(buffer, nxt::BufferUsageBit::TransferSrc) ||
    !state->ValidateUseTextureAs(...)) {
}
here non-state dependent validation.
```

```Validate``` state methods could be const, provided we are ok making aspects mutable for lazy computation of the bindgroups aspect.

Also the state tracker could be used as a local variable inside ```ValidateGetResult``` and the relevant data (such as transitioned resources) extracted at the end of the function.

**Comment by kainino0x** [comment]
@Kangz I've finally finished the full refactor of this state tracking and I think it goes along the guidelines you mentioned. PTAL!

Unfortunately there are a few changes which are hard to diff. I'll try to remember them:
* Removed VALIDATION_ASPECT_RENDER_PASS - I don't think it's necessary as it's not used in any of the 'fast-path' validation.
* Refactored the slow paths of ValidateCan{Dispatch,DrawArrays,DrawElements} - should be identical, though. DrawArrays and DrawElements each get their own fast path now.

All of the CommandBufferStateTracker public methods do their own HandleErrors and return bool.

File: src/backend/common/CommandBufferStateTracker.cpp [hunk 1]
```diff
@@ -0,0 +1,455 @@
...
+            builder->HandleError("Some bind groups are not set");
+            return false;
+        }
+        return true;
+    }
+
+    bool CommandBufferStateTracker::ValidateCanDrawArrays() {
+        // TODO(kainino@chromium.org): Check for a current render pass
+        constexpr ValidationAspects requiredAspects =
+            1 << VALIDATION_ASPECT_RENDER_PIPELINE |
```
Kangz: The VERTEX_BUFFERS aspect is missing here

src/backend/common/CommandBufferStateTracker.cpp
```diff
@@ -0,0 +1,455 @@
...
+            1 << VALIDATION_ASPECT_BIND_GROUPS;
+        if ((requiredAspects & ~aspects).none()) {
+            // Fast return-true path if everything is good
+            return true;
+        }
+
+        return ValidateCanDraw();
```
Kangz: ValidateCanDraw is a bit confusing because it is so similar to ValidateCanDrawArrays, maybe rename it, for example with RevalidateCanDraw to mean it isn't lazy

src/backend/common/CommandBufferStateTracker.cpp
```diff
@@ -0,0 +1,455 @@
...
+        }
+
+        currentRenderPass = renderPass;
+        currentFramebuffer = framebuffer;
+        currentSubpass = 0;
+        if (!BeginSubpass()) {
+            return false;
+        }
+
+        aspects.reset(VALIDATION_ASPECT_COMPUTE_PIPELINE);
```
Kangz: note: I don't know how inheritance of bindings and vertices works at renderpass start in the backing APIs, but it seems likely that we will have to reset all aspects here. This way applications will be forced to set things again in the subpass, which will remove the need to do some state-tracking in the backends.

src/backend/common/CommandBufferStateTracker.cpp
```diff
@@ -0,0 +1,455 @@
...
+                return false;
+            }
+            if (!pipeline->GetRenderPass()->IsCompatibleWith(currentRenderPass)) {
+                builder->HandleError("Pipeline is incompatible with this render pass");
+                return false;
+            }
+            aspects.set(VALIDATION_ASPECT_RENDER_PIPELINE);
+        }
+        aspects.reset(VALIDATION_ASPECT_BIND_GROUPS);
+        aspects.reset(VALIDATION_ASPECT_VERTEX_BUFFERS);
```
Kangz: Previous code was doing it, however we shouldn't reset vertex buffers here: all backing APIs keep the vertex buffers bindings on pipeline change.

src/backend/common/CommandBufferStateTracker.cpp
```diff
@@ -0,0 +1,455 @@
...
+        if (!ValidateBindGroupUsages(bindgroup)) {
+            return false;
+        }
+        bindgroupsSet.set(index);
+
+        return true;
+    }
+
+    bool CommandBufferStateTracker::SetIndexBuffer(BufferBase* buffer) {
+        auto usage = nxt::BufferUsageBit::Index;
```
Kangz: Here and for vertex buffers, we should check we are in a pipeline.

src/backend/common/CommandBufferStateTracker.cpp
Kangz: Sounds good, it makes things clearer and should be just as fast with LTO!

PR #20, opened 2017-05-24
Create a Null backend.

Kangz:
This will be useful to run validation unittests without using the GPU.


Kangz:
Thanks for the review!

PR #21, opened 2017-05-29
Windows compilation

Kangz:
This fixes Windows compilation locally. @kainino0x PTAL too! There are still a bunch of warnings on Windows that we should remove incrementally.

This still misses one fix for nxtcpp.h's ObjectBase that will be in the validation test PR.

### DIFF
--- a/generator/main.py
+++ b/generator/main.py
@@ -210,12 +210,12 @@ class PreprocessingLoader(jinja2.BaseLoader):
         for line in lines:
             # The capture in the regex adds one element per block start or end so we divide by two
             # there is also an extra line chunk corresponding to the line end, so we substract it.
-            numends = (len(self.blockend.split(line)) - 1) / 2
+            numends = (len(self.blockend.split(line)) - 1) // 2
             indentation_level -= numends
 
             result.append(self.remove_indentation(line, indentation_level))
 
-            numstarts = (len(self.blockstart.split(line)) - 1) / 2
+            numstarts = (len(self.blockstart.split(line)) - 1) // 2
             indentation_level += numstarts
 
         return '\n'.join(result)
@@ -456,6 +456,7 @@ def main():
             [os.path.abspath(args.json[0])] +
             [os.path.realpath(__file__)]
         )
+        dependencies = [dependency.replace('\\', '/') for dependency in dependencies]
         sys.stdout.write(output_separator.join(dependencies))
         return 0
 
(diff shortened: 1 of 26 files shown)


austinEng:
Okay, so I'm trying to run your changes and I'm getting an exception on `ClientQueueBuilderRelease`

I think it has something to do with how the handles/move semantics/rvalue references are being handled.

`return QueueBuilder::Acquire(result)` seems to be working correctly, but when `Device::CreateQueueBuilder()` returns, the move constructor (as expected) is called with `QueueBuilder&& other` as the result of `QueueBuilder::Acquire`.  However, the handle which it is swapped with is not `nullptr`. It's some random invalid pointer. `nxt::QueueBuilder::NxtRelease` then fails when calling `nxtProcTable::queueBuilderRelease` because the procTable is invalid.

austinEng:
Actually, should `nxt::ObjectBase<Derived, CType>::ObjectBase(ObjectBase&& other)` even release its handle? This seems like a reasonable thing to do in the move assignment operator, but in the move constructor, there is no `ObjectBase` that has previously been constructed that has a handle to release.

Kangz:
You are right, the move semantics were broken, and it didn't show up in Linux or Mac builds due to how compilers generate or not move constructions.

The fixes for this are in https://github.com/google/nxt-standalone/pull/22/commits/4d8857f5f7f1fb2766d2b46a6b88edeb58ad2853 Sorry I should have been more clear in the first comment.

Kangz:
And  you are right that the move constructor shouldn't need to release, I'll fix that.

generator/main.py
```diff
@@ -210,12 +210,12 @@ def preprocess(self, source):
         for line in lines:
             # The capture in the regex adds one element per block start or end so we divide by two
             # there is also an extra line chunk corresponding to the line end, so we substract it.
-            numends = (len(self.blockend.split(line)) - 1) / 2
+            numends = int((len(self.blockend.split(line)) - 1) / 2)
```
kainino0x: IIRC this can be done with `(len(self.blockend.split(line)) - 1) // 2` in both python 2 and 3. Do whichever you think is clearer, I don't have a strong feeling.

kainino0x:
> LGTM. Any chance we could add some more warning flags (-pedantic in particular) to help make sure that we don't accidentally break the build on other platforms?

(Doesn't have to be in this PR, just something I would really like to have now that we're developing on more than one compiler. Other warnings would be good too but probably harder to clean up. If you'd like I can try to make this switch, LMK.)

Kangz:
Thanks for the review, I'll look at setting up travis and Appveyor soon so we get compilation and unittests tested on every PR.

generator/main.py
Kangz: TIL, done

PR #22, opened 2017-05-29
Validation tests

Kangz:
PTAL, this is adds the skeleton of validation tests.

I went through many iterations to try to make this work seamlessly with an added wire between the test code and the null backend, but didn't find a good way to do it (I stopped short of generating a whole new nxtcpp just for test purpose).
If we ever want to make validation tests work with the wire, we could create an alternative to TerribleCommandBuffer that doesn't buffer at all and basically makes all nxt calls synchronous.

### DIFF
--- a/src/tests/unittests/validation/ValidationTest.cpp
+++ b/src/tests/unittests/validation/ValidationTest.cpp
@@ -1,0 +1,70 @@
+// Copyright 2017 The NXT Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


src/tests/unittests/validation/ValidationTest.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+void ValidationTest::TearDown() {
+    for (auto& expectation : expectations) {
+        // Double the gtest asserts with prints of the debug name to help with troubleshooting
+
+        if (!expectation.gotStatus) {
+            std::cerr << "Didn't get a status for " << expectation.debugName << std::endl;
+        }
+        ASSERT_TRUE(expectation.gotStatus);
```
kainino0x: I think this works:

`ASSERT_TRUE(expectation.gotStatus) << "Didn't get a status for " << expectation.debugName;`

and would be cleaner.

kainino0x:
Looks great!

Kangz:
Thanks for the review, this feature of gtest makes things much nicer!

PR #26, opened 2017-05-31
Depth and stencil testing

austinEng:
This PR adds depth and stencil testing. The `HelloDepthStencil` example shows this working on the OpenGL backend. It uses the stencil test to draw a cube's reflection on a plane. The Metal backend is completely untested, but builds on CI.

Notes:
- To my knowledge, [D3D12](https://msdn.microsoft.com/en-us/library/windows/desktop/dn903887(v=vs.85).aspx) does not support separate front and back face stencil reference values whereas [Metal](https://developer.apple.com/reference/metal/mtlrendercommandencoder/1515522-setstencilfrontreferencevalue?language=objc) and [Vulkan](https://www.khronos.org/registry/vulkan/specs/1.0/apispec.html#_vkcmdsetstencilreference_3) do. This PR currently allows two separate reference values. Should NXT only support the most limiting API?
- Vulkan allows dynamically changing the stencil mask while the other APIs do not. Should NXT support this and clone, modify, and set the depth stencil descriptor in the other APIs?


next.json
```diff
@@ -280,6 +280,13 @@
                 ]
             },
             {
+                "name": "set stencil reference",
+                "args": [
+                    {"name": "back reference", "type": "uint32_t"},
```
Kangz: Like you mentioned, D3D12 can't set the front and back stencil refs separately, so we should no advertise that option.

next.json
```diff
@@ -794,5 +912,8 @@
     },
     "uint32_t": {
         "category": "native"
+    },
+    "bool": {
```
Kangz: nit: try to keep this file sorted as much as possible.

next.json
```diff
@@ -414,9 +438,84 @@
...
+                "args": [
+                    {"name": "face", "type": "face"},
+                    {"name": "read mask", "type": "uint32_t"},
+                    {"name": "write mask", "type": "uint32_t"}
+                ]
+            }
+        ]
+    },
+    "depth write mode": {
+        "category": "enum",
```
Kangz: This could be replaced by a bool, since it is essentially a bool value in all APIs.

next.json
```diff
@@ -414,9 +438,84 @@
...
+    },
+    "depth stencil state builder": {
+        "category": "object",
+        "methods": [
+            {
+                "name": "get result",
+                "returns": "depth stencil state"
+            },
+            {
+                "name": "set depth enabled",
```
Kangz: nit: I kind of like the Metal approach of having this be implicitly defined by whether the compare function is "always" or not. What do you think?

next.json
```diff
@@ -414,9 +438,84 @@
...
+                "name": "set stencil compare function",
+                "args": [
+                    {"name": "face", "type": "face"},
+                    {"name": "stencil compare function", "type": "compare function"}
+                ]
+            },
+            {
+                "name": "set stencil mask",
+                "args": [
+                    {"name": "face", "type": "face"},
```
Kangz: Separate read and write masks are not supported in D3D12.

next.json
```diff
@@ -414,9 +438,84 @@
...
+                    {"name": "stencil failure operation", "type": "stencil operation"},
+                    {"name": "depth failure operation", "type": "stencil operation"},
+                    {"name": "stencil pass operation", "type": "stencil operation"}
+                ]
+            },
+            {
+                "name": "set stencil compare function",
+                "args": [
+                    {"name": "face", "type": "face"},
+                    {"name": "stencil compare function", "type": "compare function"}
```
Kangz: nit: could be merged with the stencil op? What do you think?

next.json
austinEng: I think this is a good idea. Do we want to do something like this for the stencil test as well? We could also infer that the stencil test will do nothing based on the stencil operation (stencilFail: Keep, depthFail: Keep, depthPass: Replace).

However for the stencil test it might be annoying to remember what values disable it so perhaps a helper function in the API to enable/disable would still be useful?

Or I suppose the stencil would default to `stencilFail: Keep, depthFail: Keep, depthPass: Replace` (off). And any changes to that would imply you wanted to enable the stencil test?

next.json
```diff
@@ -535,6 +634,12 @@
                 "args": [
                     {"name": "input", "type": "input state"}
                 ]
+            },
+            {
+                "name": "set depth stencil state",
+                "args": [
+                    {"name": "input", "type": "depth stencil state"}
```
kainino0x: nit: `"name": "state"` or something. ("input" was probably also a bad name for the argument to SetInputState, my bad)

next.json
Kangz: Also it would be nice to define a default value for this in a renderpass, so that we don't have UB when a draw needs to use the reference and it hasn't been set in the command buffer before.

examples/HelloDepthStencil.cpp
```diff
@@ -0,0 +1,352 @@
...
+        12, 13, 14,
+        12, 14, 15,
+
+        16, 17, 18,
+        16, 18, 19,
+
+        20, 21, 22,
+        20, 22, 23
+    };
+    indexBuffer = device.CreateBufferBuilder()
```
Kangz: nit: we could have a helper function in Utils that given a (device, pointer, size, final usage) creates a buffer with that single usage and data.

examples/HelloDepthStencil.cpp
```diff
@@ -0,0 +1,352 @@
...
+        void main() {
+            fragColor = vec4(mix(f_col, vec3(0.5, 0.5, 0.5), 0.5), 1.0);
+        })"
+    );
+
+    auto inputState = device.CreateInputStateBuilder()
+        .SetAttribute(0, 0, nxt::VertexFormat::FloatR32G32B32, 0)
+        .SetAttribute(1, 0, nxt::VertexFormat::FloatR32G32B32, 3 * sizeof(float))
+        .SetInput(0, 6 * sizeof(float), nxt::InputStepMode::Vertex)
+        .SetInput(1, 6 * sizeof(float), nxt::InputStepMode::Vertex)
```
Kangz: nit: you are only using bindingSlot 0 so you don't need to SetInput(1, ...)

src/backend/common/Pipeline.cpp
```diff
@@ -27,7 +28,7 @@ namespace backend {
     PipelineBase::PipelineBase(PipelineBuilder* builder)
         : device(builder->device), stageMask(builder->stageMask), layout(std::move(builder->layout)),
           renderPass(std::move(builder->renderPass)), subpass(builder->subpass),
-          inputState(std::move(builder->inputState)) {
+          inputState(std::move(builder->inputState)), depthStencilState(std::move(builder->depthStencilState)) {
```
Kangz: Can you add validation in this function that if IsCompute then we don't have non-compute state?

Kangz:
Also once you've addressed the comments, please add @kainino0x to the review. Going forward I'd like to have each PR have two reviewers.

next.json
Kangz: Both for stencil and depth there are two separate aspects: the test and the writing to the framebuffer. For the depth, ComparisonMode::Always is equivalent to disabling the depth test but not the depth write, which is controlled separately.

We should do the same for stencil: ComparisonMode::Always is equivalent to disabling the stencil test and having all stencil ops be Keep is equivalent to disabling the stencil write. These should be the default.

austinEng:
@Kangz @kainino0x Updated, pending that the Metal backend builds correctly. 
> Looking at the metal headers it looks like it takes stuff by value but can be reset with null, so we don't need to allocate backFaceStencil and frontFaceStencil.

For this I still had to allocate them because I got an error saying that these interface types could not be statically allocated.

examples/HelloDepthStencil.cpp
```diff
@@ -0,0 +1,322 @@
...
+        2.0, -1.0,  2.0,    0.5, 0.5, 0.5,
+        -2.0, -1.0,  2.0,    0.5, 0.5, 0.5,
+    };
+    planeBuffer = CreateFrozenBufferFromData(device, (void*)planeData, sizeof(planeData), nxt::BufferUsageBit::Vertex);
+}
+
+struct {
+    glm::mat4 view;
+    glm::mat4 proj;
+} cameraData;
```
Kangz: nit: in C++ put the name after the struct keyword (+ types in CamelCase).

```
struct CameraData {
    // ...
};
```

examples/HelloDepthStencil.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+    cameraBuffer = device.CreateBufferBuilder()
+        .SetAllowedUsage(nxt::BufferUsageBit::Mapped | nxt::BufferUsageBit::Uniform)
+        .SetInitialUsage(nxt::BufferUsageBit::Mapped)
+        .SetSize(sizeof(cameraData))
+        .GetResult();
+
+    glm::mat4 transform(1.0);
+
+    transformBuffer[0] = device.CreateBufferBuilder()
```
Kangz: nit: this can be created frozen too

examples/HelloDepthStencil.cpp
```diff
@@ -0,0 +1,322 @@
...
+        .SetAllowedUsage(nxt::BufferUsageBit::Mapped | nxt::BufferUsageBit::Uniform)
+        .SetInitialUsage(nxt::BufferUsageBit::Mapped)
+        .SetSize(sizeof(glm::mat4))
+        .GetResult();
+    transformBuffer[0].SetSubData(0, sizeof(glm::mat4) / sizeof(uint32_t), reinterpret_cast<uint32_t*>(&transform));
+    transformBuffer[0].FreezeUsage(nxt::BufferUsageBit::Uniform);
+
+    transform = glm::translate(transform, glm::vec3(0.f, -2.f, 0.f));
+
+    transformBuffer[1] = device.CreateBufferBuilder()
```
Kangz: ditto.

examples/Utils.h
```diff
@@ -34,7 +34,9 @@ extern "C" {
     nxt::Device CreateCppNXTDevice();
     nxt::ShaderModule CreateShaderModule(const nxt::Device& device, nxt::ShaderStage stage, const char* source);
     void CreateDefaultRenderPass(const nxt::Device& device, nxt::RenderPass* renderPass, nxt::Framebuffer* framebuffer);
+    nxt::Buffer CreateFrozenBufferFromData(const nxt::Device& device, void* data, uint32_t size, nxt::BufferUsageBit usage);
 #else
     nxtDevice CreateNXTDevice();
     nxtShaderModule CreateShaderModule(nxtDevice device, nxtShaderStage stage, const char* source);
+    nxtBuffer CreateFrozenBufferFromData(nxtDevice device, void* data, uint32_t size, nxtBufferUsageBit usage);
```
Kangz: nit: for next time I think you don't need to add the C version of the API. I'm thinking of making CHelloTriangle a C++ file that calls the C API.

next.json
Kangz: Read and write both exist in D3D but they can't be set for each face separately, sorry my comment wasn't very clear.

src/backend/CMakeLists.txt
```diff
@@ -83,6 +85,8 @@ list(APPEND BACKEND_SOURCES
     ${OPENGL_DIR}/CommandBufferGL.h
```
Kangz: nit: move nxt::opengl::DepthStencilState in its own file, DepthStencilStateGL.h/cpp

src/backend/common/CommandBuffer.cpp
```diff
@@ -564,6 +570,20 @@ namespace backend {
                     }
                     break;
 
+                case Command::SetStencilReference:
+                    {
+                        SetStencilReferenceCmd* cmd = iterator.NextCommand<SetStencilReferenceCmd>();
+                        if (lastPipeline->IsCompute()) {
```
Kangz: Compute workloads are done outside of renderpasses and graphics inside so you don't need to check for lastPipeline->IsCompute.

src/backend/common/DepthStencilState.h
```diff
@@ -0,0 +1,84 @@
...
+            struct StencilFaceInfo {
+                nxt::CompareFunction compareFunction = nxt::CompareFunction::Always;
+                nxt::StencilOperation stencilFail = nxt::StencilOperation::Keep;
+                nxt::StencilOperation depthFail = nxt::StencilOperation::Keep;
+                nxt::StencilOperation depthStencilPass = nxt::StencilOperation::Keep;
+                uint32_t mask = 0xff;
+            };
+
+            struct StencilInfo {
+                bool stencilTestEnabled;
```
Kangz: stencilEnabled is never used.
Also the read and write mask will live inside StencilInfo

src/backend/common/DepthStencilState.cpp
```diff
@@ -0,0 +1,153 @@
...
+    DepthStencilStateBase::DepthStencilStateBase(DepthStencilStateBuilder* builder)
+        : depthInfo(builder->depthInfo), stencilInfo(builder->stencilInfo) {
+    }
+
+    bool DepthStencilStateBase::DepthTestEnabled() const {
+        return depthInfo.compareFunction != nxt::CompareFunction::Always;
+    }
+
+    bool DepthStencilStateBase::StencilTestEnabled() const {
+        return (stencilInfo.back.compareFunction != nxt::CompareFunction::Always ||
```
Kangz: That's not pretty but is ok. Can you indent the continuation lines?

src/backend/common/Pipeline.cpp
```diff
@@ -118,10 +128,17 @@ namespace backend {
         if (!inputState) {
```
Kangz: Can you add the same && !IsCompute() for inputState?

src/backend/metal/MetalBackend.mm
```diff
@@ -648,6 +663,99 @@ void EndRenderPass() {
...
+            frontFaceStencil.stencilCompareFunction = MetalDepthStencilCompareFunction(stencil.front.compareFunction);
+            frontFaceStencil.stencilFailureOperation = MetalStencilOperation(stencil.front.stencilFail);
+            frontFaceStencil.depthFailureOperation = MetalStencilOperation(stencil.front.depthFail);
+            frontFaceStencil.depthStencilPassOperation = MetalStencilOperation(stencil.front.depthStencilPass);
+            frontFaceStencil.readMask = stencil.front.mask;
+            frontFaceStencil.writeMask = stencil.front.mask;
+
+            mtlDepthStencilDescriptor.backFaceStencil = backFaceStencil;
+            mtlDepthStencilDescriptor.frontFaceStencil = frontFaceStencil;
+            [backFaceStencil release];
```
Kangz: Checked that this early release doesn't cause a crash, I'm not sure what ObjC magic makes this work.

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -26,6 +27,8 @@
 namespace backend {
 namespace opengl {
 
+    PersistentPipelineState persistentPipelineState;
```
Kangz: This should go in Execute.

Kangz: Also at the beginning of Execute, we should probably reset all the states to the defaults.

src/backend/opengl/OpenGLBackend.cpp
```diff
@@ -43,6 +45,52 @@ namespace opengl {
         *device = reinterpret_cast<nxtDevice>(new Device);
     }
 
+    GLuint OpenGLCompareFunction(nxt::CompareFunction compareFunction) {
```
Kangz: nit: both of these could go in the DepthStencil file and could even be made static (or in an anonymous namespace, which is the C++ equivalent)

src/tests/unittests/validation/DepthStencilStateValidationTest.cpp
```diff
@@ -0,0 +1,82 @@
...
+// limitations under the License.
+
+#include "ValidationTest.h"
+
+class DepthStencilStateValidationTest : public ValidationTest {
+};
+
+TEST_F(DepthStencilStateValidationTest, Creation) {
+    // Success
+    nxt::DepthStencilState buf0 = AssertWillBeSuccess(device.CreateDepthStencilStateBuilder())
```
Kangz: nit: you could even do things like this so you can avoid naming each object differently.

```
{
    nxt::DepthStencilState ds = AssertWillBeSuccess(blah blah)
        .blah
        .foo
        .GetResult();
}
```

src/tests/unittests/validation/DepthStencilStateValidationTest.cpp
```diff
@@ -0,0 +1,82 @@
...
+    nxt::DepthStencilState buf1 = AssertWillBeSuccess(device.CreateDepthStencilStateBuilder())
+        .GetResult();
+
+    // Test failure when specifying properties multiple times
+    nxt::DepthStencilState buf2 = AssertWillBeError(device.CreateDepthStencilStateBuilder())
+        .SetDepthWriteEnabled(true)
+        .SetDepthWriteEnabled(false)
+        .GetResult();
+
+    nxt::DepthStencilState buf3 = AssertWillBeError(device.CreateDepthStencilStateBuilder())
```
Kangz: nit: let's have a different description for each test.

src/tests/unittests/validation/DepthStencilStateValidationTest.cpp
```diff
@@ -0,0 +1,82 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "ValidationTest.h"
+
+class DepthStencilStateValidationTest : public ValidationTest {
+};
+
+TEST_F(DepthStencilStateValidationTest, Creation) {
```
Kangz: nit: CreationDuplicates, I will update BufferValidationTest to be the same.

src/tests/unittests/validation/DepthStencilStateValidationTest.cpp
```diff
@@ -0,0 +1,82 @@
...
+        .SetStencilMask(nxt::Face::Back, 0xff)
+        .GetResult();
+
+    nxt::DepthStencilState buf9 = AssertWillBeError(device.CreateDepthStencilStateBuilder())
+        .SetStencilFunction(nxt::Face::Both, nxt::CompareFunction::Less,
+            nxt::StencilOperation::Replace, nxt::StencilOperation::Replace, nxt::StencilOperation::Replace)
+        .SetStencilFunction(nxt::Face::Back, nxt::CompareFunction::Greater,
+            nxt::StencilOperation::Replace, nxt::StencilOperation::Replace, nxt::StencilOperation::Replace)
+        .GetResult();
+}
```
Kangz: We also need to add a test for setting a depth stencil test on a compute pipeline. This is significant amount of work so feel free to do it in a follow up PR. You'll need to add a function like Utils.h CreateShaderModule to the tests and also add a dummy compute shader.

src/backend/opengl/PersistentPipelineStateGL.h
```diff
@@ -0,0 +1,72 @@
...
+                DepthStencilStateBase::DepthInfo depthInfo;
+                DepthStencilStateBase::StencilInfo stencilInfo;
+                uint32_t stencilReference;
+            };
+
+        private:
+            State state;
+            std::bitset<Field::Count> dirtyFields;
+
+            inline bool IsDirty(Field field) const;
```
Kangz: inline does nothing when the implementation isn't in a header. And also we aren't going to rely on inline, but let compilers do their optimizations in LTO instead.

src/backend/opengl/PersistentPipelineStateGL.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+            CleanField(STENCIL_FRONT_COMPARE_FUNCTION);
+            CleanField(STENCIL_FRONT_MASK);
+        }
+
+        CleanField(STENCIL_REFERENCE); // clean this last because its used for both the back and front functions
+    }
+
+}
+}
```
Kangz: I think you are missing a newline at the end of the file (which is actually against the C standard IIRC)

src/backend/opengl/PersistentPipelineStateGL.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include <bitset>
+
+namespace backend {
+namespace opengl {
+
+    class PersistentPipelineState {
+        public:
+            PersistentPipelineState();
+            void UpdateDepthStencilInfo(const DepthStencilStateBase* const depthStencilState);
```
Kangz: Ideally this would be a "dumb" struct and wouldn't extract arguments from DepthStencilState but instead DepthStencilState would pass the required arguments (I think there are only 4 of them).

src/backend/opengl/PersistentPipelineStateGL.h
```diff
@@ -0,0 +1,72 @@
...
+#define BACKEND_OPENGL_PERSISTENTPIPELINESTATE_H_
+
+#include "common/DepthStencilState.h"
+
+#include <bitset>
+
+namespace backend {
+namespace opengl {
+
+    class PersistentPipelineState {
```
Kangz: Per offline discussion this could be simplified to only by having only the following methods:

```
void SetDefaultState();
void SetStencilOpAndMasks(frontOp, backOp, readMask, writeMask);
void SetStencilReference(ref);
```

That check if the values changed and apply the state immediately if needed.

src/backend/opengl/PersistentPipelineStateGL.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    // when a field on PersistentPipelineState::State changes, mark its starting location as dirty
+    #define SET_FIELD(field, destination, value) { \
+        if (state.destination != value) { \
+            dirtyFields.set(field); \
+            state.destination = value; \
+        } \
+    }
+
+    void PersistentPipelineState::UpdateDepthStencilInfo(const DepthStencilStateBase* const depthStencilState) {
```
Kangz: nit: we generally don't need constant pointers, only pointers to constant data (and the function is free to point to a different object if it wants to). (aka you can remove the second const)

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -184,9 +185,8 @@ namespace opengl {
                 case Command::SetStencilReference:
                     {
                         SetStencilReferenceCmd* cmd = commands.NextCommand<SetStencilReferenceCmd>();
-                        DepthStencilState* depthStencilState = ToBackend(lastPipeline->GetDepthStencilState());
-                        persistentPipelineState.UpdateStencilReference(cmd->reference);
-                        persistentPipelineState.ApplyStencilNow();
+                        // DepthStencilState* depthStencilState = ToBackend(lastPipeline->GetDepthStencilState());
```
Kangz: nit: commented out code?

src/backend/opengl/DepthStencilStateGL.cpp
```diff
@@ -0,0 +1,128 @@
...
+            OpenGLStencilOperation(stencilInfo.back.stencilFail),
+            OpenGLStencilOperation(stencilInfo.back.depthFail),
+            OpenGLStencilOperation(stencilInfo.back.depthStencilPass)
+        );
+        glStencilFuncSeparate(GL_BACK,
+            backCompareFunction,
+            persistentPipelineState.GetCachedStencilReference(),
+            stencilInfo.readMask
+        );
+        glStencilMaskSeparate(GL_BACK, stencilInfo.writeMask);
```
Kangz: nit: the masks aren't separate so you could set both the back and front masks at the same time.

src/backend/opengl/DepthStencilStateGL.cpp
```diff
@@ -0,0 +1,128 @@
...
+        GLenum frontCompareFunction = OpenGLCompareFunction(stencilInfo.front.compareFunction);
+
+        persistentPipelineState.CacheStencilFuncsAndMask(backCompareFunction, frontCompareFunction, stencilInfo.readMask);
+
+        glStencilOpSeparate(GL_BACK,
+            OpenGLStencilOperation(stencilInfo.back.stencilFail),
+            OpenGLStencilOperation(stencilInfo.back.depthFail),
+            OpenGLStencilOperation(stencilInfo.back.depthStencilPass)
+        );
+        glStencilFuncSeparate(GL_BACK,
```
Kangz: I was thinking this call would be done inside a persistentPipelineState.SetStencilFuncsAndMask(); so glStencilFuncSeparate wouldn't appear here and would be only done in PersistentPipelineState.

PR #27, opened 2017-05-31
Add AppVeyor CI

Kangz:
PTAL @kainino0x @austinEng 
AppVeyor support was mostly painless but I only put it on one configuration. Ideally we will extend it to test in Release, other versions of VS and maybe different archs.


Kangz:
Thanks for the review, since this is small, landing it without the second review.

PR #31, opened 2017-06-05
Spirv cross upgrade

Kangz:
PTAL, this upgrades SPIRV-Cross so that we get HLSL support. Also fixes a couple bug.

For the Dispatch validation I wanted to add a test but validation tests don't know how to compile stuff yet so I put a TODO to make a test instead.


src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -66,12 +66,14 @@ namespace opengl {
             switch (type) {
                 case Command::AdvanceSubpass:
                     {
+                        commands.NextCommand<AdvanceSubpassCmd>();
```
kainino0x: oops, thanks for catching this.

PR #36, opened 2017-06-07
Move Resource Barriers to D3D12 Binding

austinEng:
- Moves resource barriers from the D3D12 backend to the binding. Previously, resource transitions between `D3D12_RESOURCE_STATE_RENDER_TARGET` and `D3D12_RESOURCE_STATE_PRESENT` were done at the start and end of every render pass. This is incorrect and should only be done when the render target changes and buffers are swapped.

- With the binding handling the frame transitions, each frame has it's own `ID3D12CommandAllocator` so the next frame can be started before the current frame is done. This is what Microsoft shows to do in their frame buffering [example](https://github.com/Microsoft/DirectX-Graphics-Samples/blob/master/Samples/Desktop/D3D12HelloWorld/src/HelloFrameBuffering/D3D12HelloFrameBuffering.cpp)

- Fix which prevents a `Queue` from resetting an `ID3D12CommandAllocator` before its commands have completed on the GPU.

### DIFF
--- a/examples/D3D12Binding.cpp
+++ b/examples/D3D12Binding.cpp
@@ -31,7 +31,7 @@ namespace backend {
 namespace d3d12 {
     void Init(ComPtr<ID3D12Device> d3d12Device, nxtProcTable* procs, nxtDevice* device);
     ComPtr<ID3D12CommandQueue> GetCommandQueue(nxtDevice device);
-    void SetNextRenderTarget(nxtDevice device, ComPtr<ID3D12Resource> renderTargetResource, D3D12_CPU_DESCRIPTOR_HANDLE renderTargetDescriptor);
+    void SetNextRenderTargetDescriptor(nxtDevice device, D3D12_CPU_DESCRIPTOR_HANDLE renderTargetDescriptor);
 }
 }
 
@@ -92,7 +92,6 @@ class D3D12Binding : public BackendBinding {
                 &swapChain1
             ));
             ASSERT_SUCCESS(swapChain1.As(&swapChain));
-            frameIndex = swapChain->GetCurrentBackBufferIndex();
 
             // Describe and create a render target view (RTV) descriptor heap.
             D3D12_DESCRIPTOR_HEAP_DESC rtvHeapDesc = {};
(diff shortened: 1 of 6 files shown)


src/backend/d3d12/QueueD3D12.cpp
```diff
@@ -32,9 +32,21 @@ namespace d3d12 {
...
         ));
         ASSERT_SUCCESS(commandList->Close());
+
+        ASSERT_SUCCESS(device->GetD3D12Device()->CreateFence(fenceValue, D3D12_FENCE_FLAG_NONE, IID_PPV_ARGS(&fence)));
+        fenceEvent = CreateEvent(nullptr, FALSE, FALSE, nullptr);
+        ASSERT(fenceEvent != nullptr);
     }
 
     void Queue::Submit(uint32_t numCommands, CommandBuffer* const * commands) {
+        const uint64_t currentFence = fenceValue++;
```
Kangz: nit: can you add a TODO saying that this stalls on the previous submit but we should fix it at some point?

examples/D3D12Binding.cpp
```diff
@@ -152,14 +172,47 @@ class D3D12Binding : public BackendBinding {
         ComPtr<IDXGISwapChain3> swapChain;
         ComPtr<ID3D12DescriptorHeap> renderTargetViewHeap;
         ComPtr<ID3D12Resource> renderTargetResources[kFrameCount];
-        uint64_t fenceValues[kFrameCount] = { 0, 0 };
+        ComPtr<ID3D12CommandAllocator> commandAllocators[kFrameCount];
+        ComPtr<ID3D12GraphicsCommandList> commandList;
+        uint64_t fenceValues[kFrameCount] = {0};
```
Kangz: Can you also address this comment from the other review? Especially since you introduced a framenumber.

> I found the code dealing with fenceValues hard to follow. To make it a bit simpler, you could have only one fence-updated variable, and call it something like "lastFrameFinishedSerial". You'll need another variable (not used in a fence) that's "currentFrameSerial". What do you think?

examples/D3D12Binding.cpp
```diff
@@ -103,40 +103,60 @@ class D3D12Binding : public BackendBinding {
...
             fenceEvent = CreateEvent(nullptr, FALSE, FALSE, nullptr);
             ASSERT(fenceEvent != nullptr);
 
-            SetNextRenderTarget();
+            SetNextRenderTarget(frameIndex);
         }
         void SwapBuffers() override {
-            // Present the rendered frame
+            // Execute commands to present the current frame's render target
+            PresentFrame(frameIndex);
```
Kangz: PresentFrame is only called here, either it could be inlined here, or the swapChain->Present call should be moved inside PresentFrame (and PresentFrame renamed to something like EnqueuePresentCommands).

examples/D3D12Binding.cpp
```diff
@@ -152,14 +172,47 @@ class D3D12Binding : public BackendBinding {
...
+            commandList->ResourceBarrier(1, &resourceBarrier);
+
+            ASSERT_SUCCESS(commandList->Close());
+            ID3D12CommandList* commandLists[] = { commandList.Get() };
+            commandQueue->ExecuteCommandLists(_countof(commandLists), commandLists);
+            ASSERT_SUCCESS(commandList->Reset(commandAllocators[frameIndex].Get(), nullptr));
+        }
+
+        void PresentFrame(uint32_t frameIndex) {
+            D3D12_RESOURCE_BARRIER resourceBarrier;
```
Kangz: I'm not sure what the semantics of D3D12 are, so this might not work. The code could be simplified by having SwapBuffers enqueue the following operations:

 - Transition from RENDER_TARGET to PRESENT
 - Present
 - Transition from PRESENT to RENDER_TARGET
 - Signal the fence.

Then SwapBuffers can be slightly simplified too to do the following:

 - Wait on the N -2 frame
 - Reset the allocators etc.
 - Enqueue work

src/backend/d3d12/QueueD3D12.cpp
austinEng: Done

examples/D3D12Binding.cpp
austinEng: Updated with clearer variable names

examples/D3D12Binding.cpp
austinEng: Resource transitions could not be merged, but resetting allocators was simplified

examples/D3D12Binding.cpp
austinEng: inlined

Kangz:
The commit changed indentation of most of D3D12Binding so the diff is really hard to see, can you look at it?

austinEng:
> The commit changed indentation of most of D3D12Binding so the diff is really hard to see, can you look at it?

Fixed

examples/D3D12Binding.cpp
```diff
@@ -103,65 +102,147 @@ class D3D12Binding : public BackendBinding {
...
-            ASSERT_SUCCESS(d3d12Device->CreateFence(fenceValues[frameIndex], D3D12_FENCE_FLAG_NONE, IID_PPV_ARGS(&fence)));
-            fenceValues[frameIndex]++;
+            // Get the initial render target and arbitrarily choose a "previous" render target that's different
+            previousRenderTargetIndex = renderTargetIndex = swapChain->GetCurrentBackBufferIndex();
+            previousRenderTargetIndex = renderTargetIndex == 0 ? 1 : 0;
+
+            // Initialize the current frame, the last completed frame, and all last frame each render target was used to 0
+            // so that it looks like we completed all previous frames
+            currentFrameNumber = 0;
+            for (uint32_t n = 0; n < kFrameCount; ++n) lastFrameRenderTargetWasUsed[n] = 0;
```
Kangz: nit: braceless control flow is a bit dangerous.

examples/D3D12Binding.cpp
```diff
@@ -103,65 +102,147 @@ class D3D12Binding : public BackendBinding {
...
+            {
+                ASSERT_SUCCESS(d3d12Device->CreateCommandList(
+                    0,
+                    D3D12_COMMAND_LIST_TYPE_DIRECT,
+                    commandAllocators[previousRenderTargetIndex].Get(),
+                    nullptr,
+                    IID_PPV_ARGS(&commandList)
+                ));
+
+                // Transition the first frame to be a render target
```
Kangz: nit: you can put this comment before the block

examples/D3D12Binding.cpp
```diff
@@ -103,65 +102,147 @@ class D3D12Binding : public BackendBinding {
...
+                ID3D12CommandList* commandLists[] = { commandList.Get() };
+                commandQueue->ExecuteCommandLists(_countof(commandLists), commandLists);
+            }
+
             ASSERT_SUCCESS(swapChain->Present(1, 0));
+            ASSERT_SUCCESS(commandQueue->Signal(fence.Get(), currentFrameNumber));
 
-            const uint64_t currentFenceValue = fenceValues[frameIndex];
-            ASSERT_SUCCESS(commandQueue->Signal(fence.Get(), currentFenceValue));
+            // Transition last frame's render target back to being a render target
```
Kangz: This should be done before the signal, otherwise when the fence value is updated, the commands in commandAllocators[renderTargetIndex] might still be in flight.

austinEng:
@Kangz Updated

examples/D3D12Binding.cpp
```diff
@@ -103,65 +102,150 @@ class D3D12Binding : public BackendBinding {
...
+                resourceBarrier.Transition.StateAfter = D3D12_RESOURCE_STATE_RENDER_TARGET;
+                resourceBarrier.Transition.Subresource = D3D12_RESOURCE_BARRIER_ALL_SUBRESOURCES;
+                resourceBarrier.Type = D3D12_RESOURCE_BARRIER_TYPE_TRANSITION;
+                resourceBarrier.Flags = D3D12_RESOURCE_BARRIER_FLAG_NONE;
+                commandList->ResourceBarrier(1, &resourceBarrier);
+                ASSERT_SUCCESS(commandList->Close());
+                ID3D12CommandList* commandLists[] = { commandList.Get() };
+                commandQueue->ExecuteCommandLists(_countof(commandLists), commandLists);
+            }
+            ASSERT_SUCCESS(commandQueue->Signal(fence.Get(), currentFrameNumber));
```
Kangz: The fence is signaled multiple times

austinEng:
@Kangz This was rebased against `master`. Needs approval again

Kangz:
Removed that restriction "branch must be up to date restriction", it seems too strong for now.

Issue #29, opened 2017-06-02
Add depth stencil state


Kangz:
- - [X] Implement depth stencil state
  - - [X] Make changes to next.json and state-tracking and null backend
  - - [X] Implement on the OpenGL backend
  - - [X] Implement on the Metal backend
- - [ ] Add tests for depth stencil state
  - - [ ] Add validation tests
    - - [X] For DepthStencilState creation
    - - [ ] That it cannot be used on a compute pipeline (we are probably going to split the graphics and compute pipelines types down the line)
  - - [ ] Add end2end tests (blocked on being able to have end2end tests)
- - [ ] Write documentation
  - - [ ] Add a description of the API in the NXT doc
  - - [ ] Write an investigation and explanation of why we designed the API this way.

PR #39, opened 2017-06-08
Update glTFViewer for render passes and depth stencil state

kainino0x:
#7, #29

### DIFF
--- a/examples/glTFViewer/glTFViewer.cpp
+++ b/examples/glTFViewer/glTFViewer.cpp
@@ -71,6 +71,8 @@ struct u_transform_block {
 
 nxt::Device device;
 nxt::Queue queue;
+nxt::RenderPass renderpass;
+nxt::Framebuffer framebuffer;
 
 nxt::Buffer defaultBuffer;
 std::map<std::string, nxt::Buffer> buffers;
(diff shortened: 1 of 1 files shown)


examples/glTFViewer/glTFViewer.cpp
```diff
@@ -252,10 +254,24 @@ namespace {
...
+            .SetAttachmentCount(1)
+            .AttachmentSetFormat(0, nxt::TextureFormat::R8G8B8A8Unorm)
+            .SetSubpassCount(1)
+            .SubpassSetColorAttachment(0, 0, 0)
+            .GetResult();
+        auto framebuffer = device.CreateFramebufferBuilder()
+            .SetRenderPass(renderpass)
+            // attachment 0 -> back buffer
+            // (implicit) // TODO(kainino@chromium.org): use the texture provided by WSI
+            .SetDimensions(640, 480)
```
kainino0x: It's weird to me that screen dimensions are now part of the material info. But I think it's correct?

Kangz: part of the material info? But I think it makes sense because you need DS buffers etc. Of the same size.

examples/glTFViewer/glTFViewer.cpp
```diff
@@ -496,6 +520,7 @@ namespace {
             material.uniformBuffer.SetSubData(0,
                     sizeof(u_transform_block) / sizeof(uint32_t),
                     reinterpret_cast<const uint32_t*>(&transforms));
+            cmd.BeginRenderPass(material.renderpass, material.framebuffer);
```
Kangz: This is missing an EndRenderpass, can you add validation for currentRenderpass == nullptr at the end of a CommandBuffeBuilder::ValidateGetResult(); (and ideally a test for it somewhere in the validation tests.)

kainino0x: Oops. sure.

examples/glTFViewer/glTFViewer.cpp
kainino0x: MaterialInfo is a struct I have in the glTFViewer. If we wanted to be able to resize the viewport, the materials would all have to be recompiled, basically, but that seems right. glTF 2.0 probably solves this, anyway (just one pipeline by default).

examples/glTFViewer/glTFViewer.cpp
Kangz: Thanks!

examples/glTFViewer/glTFViewer.cpp
Kangz: I'm not sure I understand: the pipeline depends on the subpass, but it is the framebuffer that has the dimension, and in all APIs, the viewport is a dynamic state.

kainino0x: Sorry, not the actual "viewport", I mean (essentially) the backbuffer size. Whatever size you want to render at.

But I think I misunderstood myself and this doesn't have to be the case. I'll figure it out.

kainino0x: Okay, I figured out why I was being silly. This is fixed now.

PR #40, opened 2017-06-09
Add validation for some null arguments and dangling render passes

kainino0x:
And add tests for both

### DIFF
--- a/src/tests/unittests/validation/CommandBufferValidationTests.cpp
+++ b/src/tests/unittests/validation/CommandBufferValidationTests.cpp
@@ -1,0 +1,65 @@
+// Copyright 2017 The NXT Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


kainino0x:
I'm pretty sure there are a bunch more null arguments checks that need to be added in many places, but this is a start.

src/tests/unittests/validation/CommandBufferValidationTests.cpp
```diff
@@ -0,0 +1,62 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "ValidationTest.h"
+
+class CommandBufferValidationTest : public ValidationTest {
+};
+
+TEST_F(CommandBufferValidationTest, Empty) {
```
Kangz: nit: comment before the tests (ANGLE-style).

Kangz:
Null arguments are not allowed anywhere in the API currently and in fact the wire doesn't support them. I'll refactor this validation to but it in the autogenerated validation in templates/BackendProcTable.cpp.

Also I'm not sure how to split test file, for now I'm thinking of splitting them depending on which builder will generate the error, and maybe have several files for the command buffer builder if needed. What do you think?

kainino0x:
null args: Okay, I was wondering about that. I've been having issues where if an invalid object is passed to a builder, it shows up on the validation side as a null pointer. I'll add comments and land, but you may want to remove these extra null checks once you add it to BackendProcTable.

split: Yes, I think that's the right split. That is (or was intended to be) how I split these files.

Kangz:
Yep the split was fine, I was just clarifying for myself :)

PR #42, opened 2017-06-09
Split Mapped usage in MapWrite and MapRead

Kangz:
Also make use of CreateFrozenBufferFromData where possible in the
example to make the renaming easier.


Kangz:
PTAL, this will make it easier to have Buffer::SetSubData() require TransferDst.

Kangz:
Thanks

Issue #8, opened 2017-05-02
Introduce buffer mapping for reading


Kangz:
In NXT buffer mapping will be the main way, if not the only one to get data back from the GPU (the other being queries). To test rendering results in automated testing we need this features. See the overview doc for the design.

 - [x] Implement buffer mapping for reading
   - [X] Implement the state tracking part
   - [X] Implement wire changes
   - [x] OpenGL backend
   - [x] Metal backend
   - [x] D3D12 backend
 - [x] Testing
   - [X] Validation tests
   - [x] end2end tests
 - Documentation
   - [X] Add a description of the API in the NXT doc
   - [X] Write an investigation and explanation of why we designed the API this way.

PR #43, opened 2017-06-12
Map read

Kangz:
PTAL! I suggest looking at commits in order, as the PR as a whole is a bit big.

Progress for #8.

### DIFF
--- a/generator/templates/wire/WireClient.cpp
+++ b/generator/templates/wire/WireClient.cpp
@@ -545,9 +545,12 @@ namespace wire {
                         }
                         buffer->mappedData = malloc(request.size);
                         memcpy(buffer->mappedData, cmd->GetData(), request.size);
+
+                        request.callback(static_cast<nxtBufferMapReadStatus>(cmd->status), buffer->mappedData, request.userdata);
+                    } else {
+                        request.callback(static_cast<nxtBufferMapReadStatus>(cmd->status), nullptr, request.userdata);
                     }
 
-                    request.callback(static_cast<nxtBufferMapReadStatus>(cmd->status), buffer->mappedData, request.userdata);
                     buffer->readRequests.erase(requestIt);
                     return true;
                 }
--- a/generator/templates/wire/WireClient.cpp
+++ b/generator/templates/wire/WireClient.cpp
@@ -21,6 +21,7 @@
 #include <vector>
 
 #include <iostream>
+#include <map>
 
 namespace nxt {
 namespace wire {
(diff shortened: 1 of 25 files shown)


generator/templates/wire/WireClient.cpp
```diff
@@ -300,11 +384,25 @@ namespace wire {
             self->errorUserdata = userdata;
         }
 
+        // XXX hanlde return command
```
austinEng: nit: hanlde --> handle. What does `XXX` mean?

src/tests/unittests/validation/BufferValidationTests.cpp
```diff
@@ -14,24 +14,270 @@
...
+        EXPECT_CALL(*mockBufferMapReadCallback, Call(NXT_BUFFER_MAP_READ_STATUS_UNKNOWN, nullptr, userdata))
+            .Times(1);
+    }
+
+    // Submitting the queue makes the null backend process map request, but the callback shouldn't
+    // be called again
+    queue.Submit(0, nullptr);
+}
+
+// Test unmapping before having the result gives UNKNOWN
```
Kangz: Done

generator/templates/wire/WireClient.cpp
Kangz: It is often used as a "FIXME", and I use it to keep lists of things to do before sending a PR. I should have removed it before committing. Also see https://www.snellman.net/blog/archive/2017-04-17-xxx-fixme/ on the lack of origin for XXX.
Done

generator/templates/wire/WireClient.cpp
```diff
@@ -411,6 +510,49 @@ namespace wire {
...
+                        }
+
+                        if (buffer->mappedData != nullptr) {
+                            return false;
+                        }
+                        buffer->mappedData = malloc(request.size);
+                        memcpy(buffer->mappedData, cmd->GetData(), request.size);
+                    }
+
+                    request.callback(static_cast<nxtBufferMapReadStatus>(cmd->status), buffer->mappedData, request.userdata);
```
Kangz: Yes failed MapReadAsync always call with nullptr, because no value would make sense in an error case. The wire makes sure requests are not confused with one another with the "request serial" that uniquely identifies a request for a given buffer.

This is also tested (briefly) in BufferValidationTest.MapReadUnmapBeforeResultThenMapAgain

austinEng: Right, but wouldn't a failed request call the callback with an error status and the previous `buffer->mappedData` which would point to valid memory if `Unmap` has not been called?

Kangz: Ah got it, fixing it by doing

```
if (success) {
    memcpy, yadi yada
    request->callback(mappedData)
} else {
    request->callback(nullptr);
}
```

Kangz: And added a test that catches the error

Kangz:
Thanks for the reviews!

PR #44, opened 2017-06-12
D3D12 vertex and index buffers with resource transitions

austinEng:
PTAL. This adds buffers to the D3D12 backend so `HelloVertices`, `HelloIndices`, and `HelloInstancing` are now working. `Buffer::SetSubData` now requires `nxt::BufferUsageBit::TransferDst`, maps an upload buffer, and schedules a copy to the backing resource. Not yet sure what to do with transitions to `MapRead` / `MapWrite` and this will likely need to be updated after #43.

There's some indentation changes in `CommandBufferD3D12.cpp` because parts of it were previously indented with 6 spaces.

### DIFF
--- a/examples/Utils.cpp
+++ b/examples/Utils.cpp
@@ -262,8 +262,8 @@ void CreateDefaultRenderPass(const nxt::Device& device, nxt::RenderPass* renderP
 
 nxt::Buffer CreateFrozenBufferFromData(const nxt::Device& device, const void* data, uint32_t size, nxt::BufferUsageBit usage) {
     nxt::Buffer buffer = device.CreateBufferBuilder()
-        .SetAllowedUsage(nxt::BufferUsageBit::MapWrite | usage)
-        .SetInitialUsage(nxt::BufferUsageBit::MapWrite)
+        .SetAllowedUsage(nxt::BufferUsageBit::TransferDst | usage)
+        .SetInitialUsage(nxt::BufferUsageBit::TransferDst)
         .SetSize(size)
         .GetResult();
     buffer.SetSubData(0, size / sizeof(uint32_t), reinterpret_cast<const uint32_t*>(data));
--- a/src/backend/common/Buffer.h
+++ b/src/backend/common/Buffer.h
@@ -35,7 +35,7 @@ namespace backend {
             bool IsTransitionPossible(nxt::BufferUsageBit usage) const;
             bool IsFrozen() const;
             bool HasFrozenUsage(nxt::BufferUsageBit usage) const;
-            void TransitionUsageImpl(nxt::BufferUsageBit usage);
+            void UpdateUsageInternal(nxt::BufferUsageBit usage);
 
             DeviceBase* GetDevice();
 
@@ -54,6 +54,7 @@ namespace backend {
             virtual void SetSubDataImpl(uint32_t start, uint32_t count, const uint32_t* data) = 0;
             virtual void MapReadAsyncImpl(uint32_t serial, uint32_t start, uint32_t size) = 0;
             virtual void UnmapImpl() = 0;
+            virtual void TransitionUsageImpl(nxt::BufferUsageBit currentUsage, nxt::BufferUsageBit targetUsage) = 0;
 
             DeviceBase* device;
             uint32_t size;
--- a/src/backend/d3d12/BufferD3D12.cpp
+++ b/src/backend/d3d12/BufferD3D12.cpp
@@ -1,0 +1,163 @@
+// Copyright 2017 The NXT Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "BufferD3D12.h"
+
+#include "D3D12Backend.h"
+
+namespace backend {
+namespace d3d12 {
+
+    namespace {
+        D3D12_RESOURCE_STATES D3D12BufferUsage(nxt::BufferUsageBit usage) {
+            D3D12_RESOURCE_STATES resourceState = D3D12_RESOURCE_STATE_COMMON;
+
+            if (usage & nxt::BufferUsageBit::TransferSrc) {
+                resourceState |= D3D12_RESOURCE_STATE_COPY_SOURCE;
+            }
+            if (usage & nxt::BufferUsageBit::TransferDst) {
+                resourceState |= D3D12_RESOURCE_STATE_COPY_DEST;
+            }
+            if (usage & (nxt::BufferUsageBit::Vertex | nxt::BufferUsageBit::Uniform)) {
+                resourceState |= D3D12_RESOURCE_STATE_VERTEX_AND_CONSTANT_BUFFER;
+            }
+            if (usage & nxt::BufferUsageBit::Index) {
+                resourceState |= D3D12_RESOURCE_STATE_INDEX_BUFFER;
+            }
+            if (usage & nxt::BufferUsageBit::Storage) {
+                resourceState |= D3D12_RESOURCE_STATE_UNORDERED_ACCESS;
+            }
+
+            return resourceState;
+        }
+    }
+
+    Buffer::Buffer(Device* device, BufferBuilder* builder)
+        : BufferBase(builder), device(device) {
+
+        D3D12_RESOURCE_DESC resourceDescriptor;
+        resourceDescriptor.Dimension = D3D12_RESOURCE_DIMENSION_BUFFER;
+        resourceDescriptor.Alignment = 0;
+        resourceDescriptor.Width = GetSize();
+        resourceDescriptor.Height = 1;
+        resourceDescriptor.DepthOrArraySize = 1;
+        resourceDescriptor.MipLevels = 1;
+        resourceDescriptor.Format = DXGI_FORMAT_UNKNOWN;
+        resourceDescriptor.SampleDesc.Count = 1;
+        resourceDescriptor.SampleDesc.Quality = 0;
+        resourceDescriptor.Layout = D3D12_TEXTURE_LAYOUT_ROW_MAJOR;
+        resourceDescriptor.Flags = D3D12_RESOURCE_FLAG_NONE;
+
+        {
+            D3D12_HEAP_PROPERTIES heapProperties;
+            heapProperties.Type = D3D12_HEAP_TYPE_UPLOAD;
+            heapProperties.CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_UNKNOWN;
+            heapProperties.MemoryPoolPreference = D3D12_MEMORY_POOL_UNKNOWN;
+            heapProperties.CreationNodeMask = 0;
+            heapProperties.VisibleNodeMask = 0;
+
+            ASSERT_SUCCESS(device->GetD3D12Device()->CreateCommittedResource(
+                &heapProperties,
+                D3D12_HEAP_FLAG_NONE,
+                &resourceDescriptor,
+                D3D12_RESOURCE_STATE_GENERIC_READ,
+                nullptr,
+                IID_PPV_ARGS(&uploadResource)
+            ));
+        }
+
+        {
+            D3D12_HEAP_PROPERTIES heapProperties;
+            heapProperties.Type = D3D12_HEAP_TYPE_DEFAULT;
+            heapProperties.CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_UNKNOWN;
+            heapProperties.MemoryPoolPreference = D3D12_MEMORY_POOL_UNKNOWN;
+            heapProperties.CreationNodeMask = 0;
+            heapProperties.VisibleNodeMask = 0;
+
+            ASSERT_SUCCESS(device->GetD3D12Device()->CreateCommittedResource(
+                &heapProperties,
+                D3D12_HEAP_FLAG_NONE,
+                &resourceDescriptor,
+                D3D12BufferUsage(GetUsage()),
+                nullptr,
+                IID_PPV_ARGS(&resource)
+            ));
+        }
+    }
+
+    ComPtr<ID3D12Resource> Buffer::GetD3D12Resource() {
+        return resource;
+    }
+
+    bool Buffer::GetResourceTransitionBarrier(nxt::BufferUsageBit currentUsage, nxt::BufferUsageBit targetUsage, D3D12_RESOURCE_BARRIER* barrier) {
+        D3D12_RESOURCE_STATES stateBefore = D3D12BufferUsage(currentUsage);
+        D3D12_RESOURCE_STATES stateAfter = D3D12BufferUsage(targetUsage);
+
+        if (stateBefore == stateAfter) {
+            return false;
+        }
+
+        barrier->Type = D3D12_RESOURCE_BARRIER_TYPE_TRANSITION;
+        barrier->Flags = D3D12_RESOURCE_BARRIER_FLAG_NONE;
+        barrier->Transition.pResource = resource.Get();
+        barrier->Transition.StateBefore = stateBefore;
+        barrier->Transition.StateAfter = stateAfter;
+        barrier->Transition.Subresource = 0;
+
+        return true;
+    }
+
+    D3D12_GPU_VIRTUAL_ADDRESS Buffer::GetVA() const {
+        return resource->GetGPUVirtualAddress();
+    }
+
+    void Buffer::SetSubDataImpl(uint32_t start, uint32_t count, const uint32_t* data) {
+        uint32_t begin = start * sizeof(uint32_t);
+        uint32_t end = (start + count) * sizeof(uint32_t);
+
+        uint8_t* mappedResource = nullptr;
+
+        D3D12_RANGE readRange;
+        readRange.Begin = 0;
+        readRange.End = 0;
+
+        ASSERT_SUCCESS(uploadResource->Map(0, &readRange, reinterpret_cast<void**>(&mappedResource)));
+        memcpy(&mappedResource[begin], data, end - begin);
+
+        D3D12_RANGE writeRange;
+        writeRange.Begin = begin;
+        writeRange.End = end;
+
+        uploadResource->Unmap(0, &writeRange);
+
+        device->GetPendingCommandList()->CopyBufferRegion(resource.Get(), begin, uploadResource.Get(), begin, end - begin);
+    }
+
+    void Buffer::MapReadAsyncImpl(uint32_t serial, uint32_t start, uint32_t count) {
+        // TODO(cwallez@chromium.org): Implement Map Read for the null backend
+    }
+
+    void Buffer::UnmapImpl() {
+        // TODO(cwallez@chromium.org): Implement Map Read for the null backend
+    }
+
+    void Buffer::TransitionUsageImpl(nxt::BufferUsageBit currentUsage, nxt::BufferUsageBit targetUsage) {
+        D3D12_RESOURCE_BARRIER barrier;
+        if (GetResourceTransitionBarrier(currentUsage, targetUsage, &barrier)) {
+            device->GetPendingCommandList()->ResourceBarrier(1, &barrier);
+        }
+    }
+
+}
+}
--- a/src/backend/d3d12/BufferD3D12.cpp
+++ b/src/backend/d3d12/BufferD3D12.cpp
@@ -59,41 +59,22 @@ namespace d3d12 {
         resourceDescriptor.Layout = D3D12_TEXTURE_LAYOUT_ROW_MAJOR;
         resourceDescriptor.Flags = D3D12_RESOURCE_FLAG_NONE;
 
-        {
-            D3D12_HEAP_PROPERTIES heapProperties;
-            heapProperties.Type = D3D12_HEAP_TYPE_UPLOAD;
-            heapProperties.CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_UNKNOWN;
-            heapProperties.MemoryPoolPreference = D3D12_MEMORY_POOL_UNKNOWN;
-            heapProperties.CreationNodeMask = 0;
-            heapProperties.VisibleNodeMask = 0;
+        D3D12_HEAP_PROPERTIES heapProperties;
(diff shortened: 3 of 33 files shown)


examples/Utils.cpp
```diff
@@ -262,8 +262,8 @@ void CreateDefaultRenderPass(const nxt::Device& device, nxt::RenderPass* renderP
 
 nxt::Buffer CreateFrozenBufferFromData(const nxt::Device& device, const void* data, uint32_t size, nxt::BufferUsageBit usage) {
     nxt::Buffer buffer = device.CreateBufferBuilder()
-        .SetAllowedUsage(nxt::BufferUsageBit::MapWrite | usage)
```
Kangz: Can you do this changes in a separate commit?
Also there are a bunch of places that were not changed to account for this (for example glTFViewer.cpp). ```git grep MapWrite``` will give a superset of the places where this is needed.

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -16,150 +16,199 @@
...
+                            draw->instanceCount,
+                            draw->firstVertex,
+                            draw->firstInstance
+                        );
+                    }
+                    break;
+
+                case Command::DrawElements:
+                    {
+                        DrawElementsCmd* draw = commands.NextCommand<DrawElementsCmd>();
```
Kangz: note to self: Command::DrawElements changed.

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -16,150 +16,199 @@
...
-              }
-          }
-      }
+                case Command::SetBindGroup:
+                    {
+                        SetBindGroupCmd* cmd = commands.NextCommand<SetBindGroupCmd>();
+                    }
+                    break;
+
+                case Command::SetIndexBuffer:
```
Kangz: note to self: SetIndexBuffer, SetVertexBuffers and transitions changed.

src/backend/d3d12/InputStateD3D12.cpp
```diff
@@ -0,0 +1,81 @@
...
+        for (size_t i = 0; i < attributesSetMask.size(); ++i) {
+            if (!attributesSetMask[i]) {
+                continue;
+            }
+
+            D3D12_INPUT_ELEMENT_DESC& inputElementDescriptor = inputElementDescriptors[count++];
+
+            const AttributeInfo& attribute = GetAttribute(i);
+
+            inputElementDescriptor.SemanticName = "TEXCOORD";
```
Kangz: Reading spriv-hlsl.cpp it seemed the SemanticName would be TEXCOORDN instead of having N in SemanticIndex. Can you check that it works if you change a vertex location (in the layout declaration in the shader) to 15?

If it works please document it here, because the D3D12 docs aren't clear.

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -16,150 +16,199 @@
...
+                    {
+                        SetIndexBufferCmd* cmd = commands.NextCommand<SetIndexBufferCmd>();
+
+                        Buffer* buffer = ToBackend(cmd->buffer.Get());
+                        D3D12_INDEX_BUFFER_VIEW bufferView;
+                        bufferView.BufferLocation = buffer->GetD3D12Resource()->GetGPUVirtualAddress() + cmd->offset;
+                        bufferView.SizeInBytes = buffer->GetSize();
```
Kangz: This is the remaining size of the buffer, so it should be ```buffer->GetSize() - cmd.offset```. Also we should add validation that cmd->offset is less than buffer->GetSize() to avoid overflows here (no need to check for the whole size of the draws for now).

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -16,150 +16,199 @@
...
+
+                        auto inputState = ToBackend(lastPipeline->GetInputState());
+
+                        std::array<D3D12_VERTEX_BUFFER_VIEW, kMaxVertexInputs> d3d12BufferViews;
+                        for (uint32_t i = 0; i < cmd->count; ++i) {
+                            auto input = inputState->GetInput(cmd->startSlot + i);
+                            Buffer* buffer = ToBackend(buffers[i].Get());
+                            d3d12BufferViews[i].BufferLocation = buffer->GetD3D12Resource()->GetGPUVirtualAddress() + offsets[i];
+                            d3d12BufferViews[i].StrideInBytes = input.stride;
+                            d3d12BufferViews[i].SizeInBytes = buffer->GetSize();
```
Kangz: same thing about offset here.

src/backend/d3d12/BufferD3D12.cpp
```diff
@@ -0,0 +1,148 @@
...
+                resourceState |= D3D12_RESOURCE_STATE_COPY_DEST;
+            }
+            if (usage & (nxt::BufferUsageBit::Vertex | nxt::BufferUsageBit::Uniform)) {
+                resourceState |= D3D12_RESOURCE_STATE_VERTEX_AND_CONSTANT_BUFFER;
+            }
+            if (usage & nxt::BufferUsageBit::Index) {
+                resourceState |= D3D12_RESOURCE_STATE_INDEX_BUFFER;
+            }
+            if (usage & nxt::BufferUsageBit::Storage) {
+                resourceState |= D3D12_RESOURCE_STATE_PIXEL_SHADER_RESOURCE | D3D12_RESOURCE_STATE_NON_PIXEL_SHADER_RESOURCE;
```
Kangz: These usage flags seem to be for sampled textures. What we call "Shader Storage Buffer" in OpenGL and Vulkan-speak are called "Unordered Access View" in D3D12, often shortened to UAV. In https://msdn.microsoft.com/en-us/library/windows/desktop/dn986744(v=vs.85).aspx you can see that this corresponds to ```D3D12_RESOURCE_STATE_UNORDERED_ACCESS```

src/backend/d3d12/BufferD3D12.cpp
```diff
@@ -0,0 +1,148 @@
...
+            }
+            if (usage & (nxt::BufferUsageBit::MapRead | nxt::BufferUsageBit::MapWrite)) {
```
Kangz: These usage flags aren't represented via resource states in D3D12. but by the heap type. So I don't think you need to do anything here.

src/backend/d3d12/BufferD3D12.h
```diff
@@ -0,0 +1,47 @@
...
+    class Buffer : public BufferBase {
+        public:
+            Buffer(Device* device, BufferBuilder* builder);
+
+            ComPtr<ID3D12Resource> GetD3D12Resource();
+            void TransitionUsageImpl(nxt::BufferUsageBit usage) override;
+
+        private:
+            Device* device;
+            ComPtr<ID3D12Resource> uploadResource;
```
Kangz: We don't want each buffer to have its own uploadResource as it will double the memory usage of each buffer. Instead there should be some sort of buffer uploader owned by the device, which buffers use to do uploads.

The code to set up uploadResource, and the code in SetSubData would be moved in that uploader.

src/backend/d3d12/BufferD3D12.cpp
```diff
@@ -0,0 +1,148 @@
...
+                D3D12_RESOURCE_STATE_GENERIC_READ,
+                nullptr,
+                IID_PPV_ARGS(&uploadResource)
+            ));
+        }
+
+        {
+            D3D12_HEAP_PROPERTIES heapProperties;
+            heapProperties.Type = D3D12_HEAP_TYPE_DEFAULT;
+            heapProperties.CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_UNKNOWN;
```
Kangz: This is a heap located on the GPU so it seems it should be D3D12_CPU_PAGE_PROPERTY_NOT_AVAILABLE

src/backend/d3d12/BufferD3D12.cpp
```diff
@@ -0,0 +1,148 @@
...
+            heapProperties.MemoryPoolPreference = D3D12_MEMORY_POOL_UNKNOWN;
+            heapProperties.CreationNodeMask = 1;
```
Kangz: nit: we only use one GPU so these two should be 0 to be consistent with the rest of the code.

src/backend/d3d12/BufferD3D12.h
```diff
@@ -0,0 +1,47 @@
...
+namespace backend {
+namespace d3d12 {
+
+    class Device;
+
+    class Buffer : public BufferBase {
+        public:
+            Buffer(Device* device, BufferBuilder* builder);
+
+            ComPtr<ID3D12Resource> GetD3D12Resource();
```
Kangz: In most cases we want the GPU VA for we could have a GetVA() function and use it instead of GetD3D12Resource()->GetGPUVirtualAddress(). Also this could be const.

src/backend/d3d12/BufferD3D12.cpp
```diff
@@ -0,0 +1,148 @@
...
+                D3D12_HEAP_FLAG_NONE,
+                &resourceDescriptor,
+                D3D12_RESOURCE_STATE_GENERIC_READ,
+                nullptr,
+                IID_PPV_ARGS(&uploadResource)
+            ));
+        }
+
+        {
+            D3D12_HEAP_PROPERTIES heapProperties;
```
Kangz: This isn't necessary here, because MapRead and MapWrite aren't used but the heap properties would depend on whether these flags are set.

src/backend/common/Buffer.h
```diff
@@ -34,7 +34,7 @@ namespace backend {
             bool IsTransitionPossible(nxt::BufferUsageBit usage) const;
             bool IsFrozen() const;
             bool HasFrozenUsage(nxt::BufferUsageBit usage) const;
-            void TransitionUsageImpl(nxt::BufferUsageBit usage);
+            virtual void TransitionUsageImpl(nxt::BufferUsageBit usage);
```
Kangz: In a patch before the D3D12 part, let's change TransitionUsageImpl do ```virtual void TransitionUsageImpl(currentUsage, targetUsage) = 0```. TransitionUsageImpl is used in a couple places: 

 - ```BufferBase::TransitionUsage``` to change usage, this would instead call the Impl and then set the current usage.
 - Backend command buffers, this is because if there is a transition in the command buffers, this should be reflected in the current usage after queue->execute. You can introduce a ```UpdateUsageInternal``` to buffer that will just asser TranstitionPossible and update the member.
 - FreezeUsage (in this PR). It would do the same as in the first bullet.

The same change should be reflected in TextureBase.

src/backend/d3d12/QueueD3D12.cpp
```diff
@@ -31,18 +31,29 @@ namespace d3d12 {
...
-        ASSERT_SUCCESS(commandList->Close());
 
         ASSERT_SUCCESS(device->GetD3D12Device()->CreateFence(fenceValue, D3D12_FENCE_FLAG_NONE, IID_PPV_ARGS(&fence)));
         fenceEvent = CreateEvent(nullptr, FALSE, FALSE, nullptr);
         ASSERT(fenceEvent != nullptr);
     }
 
     void Queue::Submit(uint32_t numCommands, CommandBuffer* const * commands) {
-        // TODO(enga@google.com): This will stall on the previous submit because
+        ComPtr<ID3D12CommandAllocator> pendingCommandAllocator = device->GetPendingCommandAllocator();
```
Kangz: LGTM, but this will need to be revisited to not stall and properly track objects that are in flight.

src/backend/d3d12/BufferD3D12.h
Kangz: Because the Queue stall after the submit, you could simplify the uploader by linearly putting stuff in a large buffer and receiving a signal after the Queue knows GPU execution is done.

src/backend/d3d12/BufferD3D12.cpp
austinEng: I would think so too, but according to the documentation:
> When Type is not D3D12_HEAP_TYPE_CUSTOM, CPUPageProperty and MemoryPoolPreference must be ..._UNKNOWN.

I think those flags are only supposed to be used if you're creating a custom heap. Using `D3D12_CPU_PAGE_PROPERTY_NOT_AVAILABLE` throws an exception:

> D3D12 ERROR: ID3D12Device::CreateCommittedResource: D3D12_HEAP_PROPERTIES is invalid. D3D12_CPU_PAGE_PROPERTY_UNKNOWN must be used when creating heaps that are not D3D12_HEAP_TYPE_CUSTOM. [ STATE_CREATION ERROR #638: CREATERESOURCEANDHEAP_INVALIDHEAPPROPERTIES]

Kangz: > > When Type is not D3D12_HEAP_TYPE_CUSTOM, CPUPageProperty and MemoryPoolPreference must be ..._UNKNOWN.

Thanks for the explanation, I hadn't seen that.

src/backend/d3d12/InputStateD3D12.cpp
austinEng: Yea the docs are misleading. It does in fact work with the declaration as 15 (although `SemanticIndex` should be `i` and not `attribute.bindingSlot`)

examples/Utils.cpp
austinEng: Done

src/backend/d3d12/BufferD3D12.cpp
austinEng: Done

src/backend/d3d12/BufferD3D12.cpp
austinEng: Removed

src/backend/d3d12/BufferD3D12.cpp
austinEng: Done

src/backend/d3d12/BufferD3D12.h
austinEng: Done

src/backend/common/Buffer.h
austinEng: Done

src/backend/d3d12/CommandBufferD3D12.cpp
austinEng: Done

src/backend/d3d12/CommandBufferD3D12.cpp
austinEng: Done

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -16,150 +16,195 @@
...
+                            d3d12BufferViews[i].SizeInBytes = buffer->GetSize() - offsets[i];
+                        }
+
+                        commandList->IASetVertexBuffers(cmd->startSlot, cmd->count, d3d12BufferViews.data());
+                    }
+                    break;
+
+                case Command::TransitionBufferUsage:
+                    {
+                        TransitionBufferUsageCmd* cmd = commands.NextCommand<TransitionBufferUsageCmd>();
```
Kangz: This should enqueue a transition command in the current command list. You can add a helper function in the Buffer that given a command list enqueues a transition and call it both here and in d3d12::Buffer::TransitionUsageImpl. Or even better make it return the D3D12_RESOURCE_BARRIER so that it doesn't have side effects.

src/backend/d3d12/ResourceUploader.cpp
```diff
@@ -0,0 +1,90 @@
...
+        ComPtr<ID3D12Resource> uploadResource;
+
+        D3D12_HEAP_PROPERTIES heapProperties;
+        heapProperties.Type = D3D12_HEAP_TYPE_UPLOAD;
+        heapProperties.CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_UNKNOWN;
+        heapProperties.MemoryPoolPreference = D3D12_MEMORY_POOL_UNKNOWN;
+        heapProperties.CreationNodeMask = 0;
+        heapProperties.VisibleNodeMask = 0;
+
+        // TODO(enga@google.com): Allocate a large upload heap on initialization and create placed resources
```
Kangz: Even better, create a large buffer and use it as a ring upload buffer.

src/backend/d3d12/ResourceUploader.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "D3D12Backend.h"
+
+namespace backend {
+namespace d3d12 {
+
+    ResourceUploader::ResourceUploader(Device* device) : device(device) {
+        ASSERT_SUCCESS(device->GetD3D12Device()->CreateFence(uploadSerial, D3D12_FENCE_FLAG_NONE, IID_PPV_ARGS(&fence)));
+    }
+
+    void ResourceUploader::UploadResource(ComPtr<ID3D12Resource> resource, uint32_t start, uint32_t count, const uint32_t* data) {
```
Kangz: nit: can you make this API start and size arguments be in bytes and not uint32_t? The reason why the SetSubData call arguments are in uint32_t is because it makes the wire know how to copy it but should be changed eventually.

src/backend/d3d12/ResourceUploader.cpp
```diff
@@ -0,0 +1,90 @@
...
+        D3D12_RESOURCE_DESC resourceDescriptor = resource->GetDesc();
```
Kangz: Can you just create your own resource descriptor of the right size here? (or do a ringbuffer)

src/backend/d3d12/ResourceUploader.h
```diff
@@ -0,0 +1,50 @@
...
+            void SignalUploadStart();
+            void FreeCompletedResources();
+
+        private:
+            Device* device;
+
+            ComPtr<ID3D12Fence> fence;
+            uint64_t uploadSerial = 0;
+            std::vector<ComPtr<ID3D12Resource>> pendingResources;
+            std::multimap<uint64_t, ComPtr<ID3D12Resource>> uploadMap;
```
Kangz: Things tracking what's in flight could be std::vector<std::pair<serial, std::vector<ThingInFlight>>>.

src/backend/d3d12/ResourceUploader.cpp
```diff
@@ -0,0 +1,90 @@
...
+        device->GetPendingCommandList()->CopyBufferRegion(resource.Get(), begin, uploadResource.Get(), begin, end - begin);
+
+        pendingResources.push_back(uploadResource);
+    }
+
+    bool ResourceUploader::HasPendingResources() const {
+        return pendingResources.size() > 0;
+    }
+
+    void ResourceUploader::SignalUploadStart() {
```
Kangz: Ideally everything uses the same fence serial and "Flight tracking" facilities.

src/backend/d3d12/CommandBufferD3D12.cpp
austinEng: Done

src/backend/d3d12/ResourceUploader.cpp
austinEng: Done

src/backend/d3d12/ResourceUploader.cpp
austinEng: What do you mean by this? This is being done because there is an existing resource on the GPU which we're uploading to. The resource created in the upload heap to copy from needs to have the same descriptor.

src/backend/d3d12/ResourceUploader.cpp
austinEng: Done

src/backend/d3d12/ResourceUploader.h
austinEng: Done

src/backend/d3d12/BufferD3D12.h
austinEng: Done

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -16,156 +16,207 @@
...
+
+                        Buffer* buffer = ToBackend(cmd->buffer.Get());
+
+                        D3D12_RESOURCE_BARRIER barrier;
+                        if (buffer->GetResourceTransitionBarrier(buffer->GetUsage(), cmd->usage, &barrier)) {
+                            commandList->ResourceBarrier(1, &barrier);
+                        }
```
Kangz: The should call UpdateUsageInternal, same for TransitionTextureUsage below.

src/backend/d3d12/D3D12Backend.cpp
```diff
@@ -87,10 +91,54 @@ namespace d3d12 {
...
 
+    ResourceUploader* Device::GetResourceUploader() {
+        return &resourceUploader;
+    }
+
     void Device::SetNextRenderTargetDescriptor(D3D12_CPU_DESCRIPTOR_HANDLE renderTargetDescriptor) {
         this->renderTargetDescriptor = renderTargetDescriptor;
     }
 
+    void Device::tick() {
```
Kangz: nit: Tick

src/backend/d3d12/ResourceUploader.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include "D3D12Backend.h"
+
+namespace backend {
+namespace d3d12 {
+
+    ResourceUploader::ResourceUploader(Device* device) : device(device) {
+    }
+
+    void ResourceUploader::UploadToResource(ComPtr<ID3D12Resource> resource, uint32_t start, uint32_t count, const uint8_t* data) {
```
Kangz: nit: The only resource we will upload to is a buffer, so the name could reflect that.

src/backend/d3d12/ResourceUploader.cpp
Kangz: This function is using CopyBufferRegion, so it requires both resources to be buffers and you can easily make your own resource descriptor.

The comment was about this function creating a full-sized buffer on the upload heap: if ```resource``` is several megabytes and we want to update a couple bytes, it would be more efficient to just create a ```uploadResource``` of a couple bytes. Instead right now, the function does an allocation of a several megabytes and uses just a couple byte of it.

Basically uploadResource needs to have a size of ```count``` and CopyBufferRegion's fourth argument becomes 0. (and the map + memcpy logic needs to be updated too)

src/backend/d3d12/CommandBufferD3D12.cpp
austinEng: Done

src/backend/d3d12/D3D12Backend.cpp
austinEng: Done

src/backend/d3d12/ResourceUploader.cpp
austinEng: Done

src/backend/d3d12/ResourceUploader.cpp
austinEng: Done

Issue #45, opened 2017-06-12
Buffer <-> Buffer copies


Kangz:
This is needed to make anything useful with ```Buffer::MapReadAsync``` as MapRead buffers can only be TransferSrc too.

 - [x] Implement B2B copies
   - [x] State tracking changes
   - [x] OpenGL backend
   - [x] Metal backend
   - [x] D3D12 backend
 - [x] Test B2B copies
   - [x] Validation tests
   - [x] End2end tests
 - [ ] Write documentation
   - [ ] Add a description of the API in the NXT doc.
   - [ ] Write an investigation doc and explanation of why the API looks this way.

PR #46, opened 2017-06-12
B2b copy

Kangz:
PTAL, this implements (but doesn't end2end test) the buffer to buffer copies that will be necessary to copy data from the GPU to a MapRead buffer.

Part of #45


Kangz:
Thanks!

PR #48, opened 2017-06-15
Introduce SerialQueue to track in flight resources

Kangz:
PTAL


Kangz:
Do you mind if I remove the death test? We don't have release asserts for now so the tests fail in Release.

austinEng:
That's fine with me

PR #50, opened 2017-06-19
Utils split

Kangz:
PTAL, this moves a bunch of code around to make it possible to share utilities between the samples and the test suite.


src/utils/NXTHelpers.cpp
```diff
@@ -0,0 +1,106 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "NXTHelpers.h"
+
+#include <shaderc/shaderc.hpp>
+
```
austinEng: Missing `#include <cstring>` for `strlen`?

examples/CMakeLists.txt
```diff
@@ -12,74 +12,56 @@
...
-
-add_library(utils STATIC ${UTILS_SOURCES})
-target_link_libraries(utils nxt_backend nxt_wire shaderc nxtcpp nxt)
-SetCXX14(utils)
-
-add_executable(CHelloTriangle HelloTriangle.c)
-target_link_libraries(CHelloTriangle utils)
+add_executable(CHelloTriangle CHelloTriangle.cpp)
+target_link_libraries(CHelloTriangle sample_utils)
+SetCXX14(CHelloTriangle)
```
austinEng: Should this still be named `CHelloTriangle `?

src/utils/NXTHelpers.cpp
Kangz: Done

examples/CMakeLists.txt
Kangz: It is testing the C API so I think the name is still ok.

Kangz:
Thanks!

PR #51, opened 2017-06-19
Test SerialQueue::FirstSerial

austinEng:
Adds a test for SerialQueue::FirstSerial

I also noticed that SerialQueue::Enque on empty vectors would break things (and wouldn't really make sense anyway) so I added an assertion for that case.

### DIFF
--- a/src/backend/common/SerialQueue.h
+++ b/src/backend/common/SerialQueue.h
@@ -110,12 +110,14 @@ namespace backend {
 
     template<typename T>
     void SerialQueue<T>::Enqueue(const std::vector<T>& values, Serial serial) {
+        ASSERT(values.size() > 0);
         ASSERT(Empty() || storage.back().first <= serial);
         storage.emplace_back(SerialPair(serial, {values}));
     }
 
     template<typename T>
     void SerialQueue<T>::Enqueue(std::vector<T>&& values, Serial serial) {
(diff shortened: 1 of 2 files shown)


src/backend/common/SerialQueue.h
```diff
@@ -110,12 +110,14 @@ namespace backend {
 
     template<typename T>
     void SerialQueue<T>::Enqueue(const std::vector<T>& values, Serial serial) {
+        ASSERT(values.size() > 0);
```
Kangz: Why this assert? It is valid to create an empty queue and Enqueue to it.

austinEng: Hm, I thought it would be strange to call Enqueue but not actually insert anything into the queue. You would be associating nothing with some serial value.

But regardless it breaks if you enqueue an empty vector because in `T& SerialQueue<T>::Iterator::operator*() const`
```
if (serialIterator == nullptr) {
    return *storageIterator->second.begin();
}
```
`storageIterator->second` is empty.

We could just modify the dereference, or even more simply just do nothing if you enqueue an empty vector?

Kangz: Ah sorry I misread and didn't realize in was for the values that were enqueued. ASSERT sounds good now :)

PR #52, opened 2017-06-21
Basic end2end test

Kangz:
PTAL, this makes it all the way to a basic end2end test. I suggest reviewing the commits individually, especially the Metal ones, as the MetalBackend files were split in per-object files.

If that's not convenient to review, I can split this in multiple PRs.


src/backend/metal/MetalBackend.mm
```diff
@@ -179,685 +200,67 @@ void Present(nxtDevice device) {
...
-                render = nil;
-            }
-        };
-
+    uint64_t Device::GetPendingCommandSerial() {
+        // If this is called, then it means some piece of code somewhere will wait for this serial to
+        // complete. Make sure the pending command buffer is created so that it is on the worst case
+        // enqueued on the next Tick() and eventually increments the serial. Otherwise if no GPU work
+        // happens we could be waiting for this serial forever.
+        GetPendingCommandBuffer();
```
austinEng: Is this necessary? Isn't any user of the pending command list going to have to call `GetPendingCommandBuffer()` anyway?

src/backend/metal/MetalBackend.h
```diff
@@ -123,17 +109,33 @@ namespace metal {
             id<MTLTexture> GetCurrentTexture();
             id<MTLTexture> GetCurrentDepthTexture();
 
+            id<MTLCommandBuffer> GetPendingCommandBuffer();
+            void SubmitPendingCommandBuffer();
+            Serial GetPendingCommandSerial();
+            Serial GetFinishedCommandSerial() const;
```
austinEng: This function is never used because `finishedCommandSerial` is a member variable which is passed to the other `Tick` functions. Are there situations where other classes will need to query the finished Serial outside of a Tick function?

src/backend/metal/MetalBackend.mm
```diff
@@ -102,19 +118,24 @@ void Present(nxtDevice device) {
...
+        return new Texture(builder);
     }
     TextureViewBase* Device::CreateTextureView(TextureViewBuilder* builder) {
-        return new TextureView(this, builder);
+        return new TextureView(builder);
     }
 
     void Device::TickImpl() {
+        resourceUploader->Tick(finishedCommandSerial);
+
```
austinEng: Should `mapReadRequestTracker->Tick` be called here as well?

src/tests/NXTTest.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    nxtSetProcs(&backendProcs);
+    device = nxt::Device::Acquire(backendDevice);
+    queue = device.CreateQueueBuilder().GetResult();
+
+    device.SetErrorCallback(DeviceErrorCauseTestFailure, 0);
+}
+
+void NXTTest::TearDown() {
+    binding->SwapBuffers();
```
austinEng: Why is it necessary to swap buffers? Maybe add a comment

src/backend/metal/MetalBackend.mm
```diff
@@ -102,19 +118,24 @@ void Present(nxtDevice device) {
...
-        return new Texture(this, builder);
+        return new Texture(builder);
     }
     TextureViewBase* Device::CreateTextureView(TextureViewBuilder* builder) {
-        return new TextureView(this, builder);
+        return new TextureView(builder);
     }
 
     void Device::TickImpl() {
+        resourceUploader->Tick(finishedCommandSerial);
```
austinEng: note to self: In the D3D12 backend, I have been given each of these tracker classes a `Tick` function which is named explicitly to communicate what it does like `FreeCompletedAllocators`. With multiple trackers, maybe we should give them all the same interface and require Tick to be implemented.

src/backend/metal/MetalBackend.h
Kangz: I can't think of a case where this would be needed right now, removed.

src/backend/metal/MetalBackend.mm
Kangz: About naming these methods Tick, the idea is that the code outside of the class doesn't need to know whether it will free or recycle or even leak allocators. We could make it an interface method, but imho we don't have enough of these to warrant putting them in a vector and adding an interface.

src/backend/metal/MetalBackend.mm
Kangz: Good catch! It got eaten by the rebase -i. I should have ran the test after it, because right now they just hang. Done

src/backend/metal/MetalBackend.mm
Kangz: The problem was that the end2end tests defer expectations and mapping the buffers for reading until the last graphic commands have been enqueued. Mapping buffers doesn't use the pending command buffer, but still needs to know when the GPU is done executing the current commands.

The idea is that if anything asks what it the pending command serial, then it will certainly watch for it to be finished, and even if we don't have any work to send to the GPU, we should still put the signal to update the serial.

src/backend/metal/ResourceUploader.h
```diff
@@ -0,0 +1,44 @@
...
+namespace metal {
+
+    class Device;
+
+    class ResourceUploader {
+        public:
+            ResourceUploader(Device* device);
+            ~ResourceUploader();
+
+            void BufferSubData(id<MTLBuffer> buffer, uint32_t start, uint32_t size, const void* data);
```
Kangz: Ack

src/tests/NXTTest.cpp
Kangz: This is something we do in ANGLE so that you can see the graphics rendering flashing on the window to try to understand what's going on. Removed.

Kangz:
Thanks for your review!

kainino0x:
Nice cleanup of the mutexes as well. LGTM!

Kangz:
Thanks :)

Issue #53, opened 2017-06-26
Texture -> Buffer copies


Kangz:
This is needed to test pixel values in the end2end tests.

 - [x] Implement T2B copies
   - [x] State tracking changes
   - [x] OpenGL backend
   - [x] Metal backend
   - [x] D3D12 backend
 - [x] Test T2B copies
   - [x] Validation tests
   - [x] End2end tests
 - [ ] Write documentation
   - [ ] Add a description of the API in the NXT doc.
   - [ ] Write an investigation doc and explanation of why the API looks this way.

PR #54, opened 2017-06-26
Texture to buffer copies

Kangz:
PTAL, this will be important to be able to check pixel values in the end2end tests. I didn't actually test that the Metal implementation works though.

Part of #53.

### DIFF
--- a/src/backend/common/CommandBuffer.cpp
+++ b/src/backend/common/CommandBuffer.cpp
@@ -250,11 +250,6 @@ namespace backend {
                         uint64_t z = copy->z;
                         uint32_t level = copy->level;
 
-                        if (width == 0 || height == 0 || depth == 0) {
-                            HandleError("Empty copy");
-                            return false;
-                        }
-
                         // TODO(cwallez@chromium.org): check for overflows
                         uint64_t pixelSize = TextureFormatPixelSize(texture->GetFormat());
                         uint64_t dataSize = width * height * depth * pixelSize;
--- a/src/backend/common/CommandBuffer.cpp
+++ b/src/backend/common/CommandBuffer.cpp
@@ -129,6 +129,12 @@ namespace backend {
                         copy->~CopyBufferToTextureCmd();
                     }
                     break;
+                case Command::CopyTextureToBuffer:
+                    {
+                        CopyTextureToBufferCmd* copy = commands->NextCommand<CopyTextureToBufferCmd>();
+                        copy->~CopyTextureToBufferCmd();
+                    }
+                    break;
                 case Command::Dispatch:
                     {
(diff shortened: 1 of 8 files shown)


src/backend/common/CommandBuffer.cpp
```diff
@@ -29,6 +29,56 @@
...
+
+            return true;
+        }
+
+        bool FitsInBuffer(const BufferBase* buffer, uint32_t offset, uint32_t size) {
+            uint32_t bufferSize = buffer->GetSize();
+            return offset <= bufferSize && (size <= (bufferSize - offset));
+        }
+
+        bool ValidateCopySizeFitsInBuffer(CommandBufferBuilder* builder, const BufferCopyLocation& location, uint32_t dataSize) {
```
kainino0x: just an idea, it would be really nice to have a macro for `__attribute__((warn_unused_result))` on compilers which support it (clang and gcc I think)

src/backend/common/CommandBuffer.cpp
```diff
@@ -239,36 +273,14 @@ namespace backend {
...
-                        if (!state->ValidateCanCopy() ||
-                            !state->ValidateCanUseBufferAs(buffer, nxt::BufferUsageBit::TransferSrc) ||
-                            !state->ValidateCanUseTextureAs(texture, nxt::TextureUsageBit::TransferDst)) {
+                        uint32_t bufferCopySize = 0;
+                        if (!ComputeTextureCopyBufferSize(this, copy->destination, &bufferCopySize) ||
+                            !ValidateCopyLocationFitsInTexture(this, copy->destination) ||
+                            !ValidateCopySizeFitsInBuffer(this, copy->source, bufferCopySize) ||
+                            !state->ValidateCanCopy() ||
+                            !state->ValidateCanUseBufferAs(copy->source.buffer.Get(), nxt::BufferUsageBit::TransferSrc) ||
+                            !state->ValidateCanUseTextureAs(copy->destination.texture.Get(), nxt::TextureUsageBit::TransferDst)) {
```
kainino0x: great refactor!

src/backend/common/CommandBuffer.cpp
Kangz: Yes that would be amazing, and we need more compiler-stuff infrastructure for import/export symbols, defines for platforms and available backends, etc. Also this would require splitting code in common/ between common utility code (Math, compile flags, SerialQueue?, ...) and the state tracking frontend.

src/backend/common/CommandBuffer.cpp
Kangz: Thank you!

Kangz:
Thanks all for the reviews

PR #55, opened 2017-06-27
Some cleanups found when testing warnings.

Kangz:
This is to trigger a travis build, please do not land.

### DIFF
--- a/src/backend/metal/CommandBufferMTL.mm
+++ b/src/backend/metal/CommandBufferMTL.mm
@@ -119,8 +119,9 @@ namespace metal {
         };
     }
 
-    CommandBuffer::CommandBuffer(Device* device, CommandBufferBuilder* builder)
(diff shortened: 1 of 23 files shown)


Kangz:
PTAL!

src/backend/metal/CommandBufferMTL.mm
```diff
@@ -274,10 +273,6 @@ void EndRenderPass() {
                 case Command::SetPushConstants:
                     {
                         SetPushConstantsCmd* cmd = commands.NextCommand<SetPushConstantsCmd>();
-                        uint32_t* valuesUInt = commands.NextData<uint32_t>(cmd->count);
```
kainino0x: Isn't this `commands.NextData` call necessary even if the result isn't used?

Kangz: Good catch, thanks

Kangz:
Thanks!

PR #60, opened 2017-07-03
Fix shaderc not finding python when inside VS

Kangz:
Visual Studio 2017 is able to build CMake-based projects directly be
opening the folder containing the CMakeLists.txt. However when doing
this shaderc is not able to find the Python executable (it uses
find_program instead of the special Python CMake module). Help shaderc
by setting the PYTHON_EXE variable before including its CMakeLists.txt


Kangz:
Thanks

PR #62, opened 2017-07-05
Input state tests

Kangz:
PTAL, this contain the first end2end tests actually rendering something.


Kangz:
Damn that texture copy limitation is pretty annoying. We should reach out to Microsoft to get clarification about it. Feel free to make NXTTest's ReadPixels buffer bigger to make that function work.

Thank you for the reviews too!

PR #63, opened 2017-07-06
null backend: fix+test resource usage after cmdbuf transition

kainino0x:


kainino0x:
This is an issue I've been thinking of and forgetting about for months now.

This is just a test for now, because I want to discuss what the actual behavior should be.

kainino0x:
@Kangz I added some code in the null backend which should allow this test to pass. But instead, I'm getting a segfault. Something I don't understand is happening in `CommandAllocator`: `currentBlock` and `currentPtr` are getting garbage values on one iteration of NextCommandId. Not sure how this happens; I don't see anywhere `currentBlock` could get assigned garbage, and `-fsanitizer=address -fstack-protector-all` isn't finding anything before the crash.

src/backend/null/NullBackend.cpp
```diff
@@ -155,6 +157,89 @@ namespace null {
...
+
+    CommandBuffer::~CommandBuffer() {
+        FreeCommands(&commands);
+    }
+
+    void CommandBuffer::Execute() {
+        Command type;
+        while (commands.NextCommandId(&type)) {
+            switch (type) {
+                case Command::AdvanceSubpass:
```
Kangz: nit: use the new and handy SkipCommands for everything except the transitions

src/backend/null/NullBackend.cpp
```diff
@@ -155,6 +157,89 @@ namespace null {
     void Buffer::TransitionUsageImpl(nxt::BufferUsageBit currentUsage, nxt::BufferUsageBit targetUsage) {
     }
 
+    // CommandBuffer
+
+    CommandBuffer::CommandBuffer(CommandBufferBuilder* builder)
+        : CommandBufferBase(builder), device(ToBackend(builder->GetDevice())),
```
Kangz: nit: the private device member isn't used.

src/tests/unittests/validation/BufferValidationTests.cpp
```diff
@@ -317,3 +317,29 @@ TEST_F(BufferValidationTest, SetSubDataWrongUsage) {
     uint32_t foo = 0;
     ASSERT_DEVICE_ERROR(buf.SetSubData(0, 1, &foo));
 }
+
+// Test that command buffer submit changes buffer usage
```
Kangz: nit: Given all the tiny details of usage validation, I suggest we make a UsageValidationTests file.

src/backend/null/NullBackend.cpp
kainino0x: Done

src/tests/unittests/validation/BufferValidationTests.cpp
kainino0x: Done

src/backend/null/NullBackend.cpp
kainino0x: Done

PR #64, opened 2017-07-06
replace AdvanceSubpass with Begin/EndRenderSubpass

kainino0x:
This change is motivated by usage transitions: if there are usage transitions in between subpasses, AdvanceSubpass makes it unnatural. Plus, if we go the route of all-explicit attachment transitions, they have to be done between subpasses.

TODO:

* [x] Should the first BeginRenderSubpass and the last EndRenderSubpass be implicit? A little asymmetrical, but it doesn't seem like a problem. - decided not to do this for now
* [x] Fix D3D12 code (still a no-op on D3D12 backend)

### DIFF
--- a/examples/Animometer.cpp
+++ b/examples/Animometer.cpp
@@ -133,6 +133,7 @@ void frame() {
 
         nxt::CommandBufferBuilder builder = device.CreateCommandBufferBuilder()
             .BeginRenderPass(renderpass, framebuffer)
+            .BeginRenderSubpass()
             .SetPipeline(pipeline)
             .Clone();
 
@@ -144,6 +145,7 @@ void frame() {
             i++;
         }
 
+        builder.EndRenderSubpass();
         builder.EndRenderPass();
         commands[j] = builder.GetResult();
     }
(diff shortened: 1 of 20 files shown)


src/backend/common/CommandBufferStateTracker.cpp
```diff
@@ -164,10 +164,6 @@ namespace backend {
     };
```
Kangz: A new validation aspect should be added "SUBPASS" which should gate most operations that are currently gated on "RENDERPASS"

Kangz:
Looks good with the validation changes.

src/backend/common/CommandBufferStateTracker.cpp
kainino0x: Actually currently most things are gated on RENDER_PIPELINE, we don't have a RENDERPASS validation aspect. I'm not sure we need a SUBPASS validation aspect (at least, not right at the moment) - I'm not spotting any area where the current validation is wrong.

kainino0x: (subpassActive could probably be replaced with an aspect though)

Kangz: Makes thanks, LGTM

kainino0x:
Replaced subpassActive with a validation aspect (and rebased). PTAL and merge if ready.

PR #65, opened 2017-07-06
Split backend/common in backend/ and common/

Kangz:
This directory used to contain both the state tracking code for the
backends, and the common utilities that could be used both by the
backends and the rest of the code. Things are now:

 - src/common is utility code for the whole repo
 - src/backend contains libNXT's code
 - src/utils is utility code that we don't want in libNXT

This commit also changes all includes to use global paths from src/
bacause it had to touch a bunch of #include statements anyway.


kainino0x:
LGTM once it's building. (looks like it's currently not.)

Kangz:
Had forgotten to add a couple files.

PR #67, opened 2017-07-07
Freeze texture transitions while attached

kainino0x:
Also changes the attachment usages to OutputAttachment and InputAttachment instead of ColorAttachment, DepthStencilAttachment, etc.

~~Based on top of #64, so there's an extra commit in here.~~

### DIFF
--- a/next.json
+++ b/next.json
@@ -921,8 +921,8 @@
             {"value": 2, "name": "transfer dst"},
             {"value": 4, "name": "sampled"},
             {"value": 8, "name": "storage"},
-            {"value": 16, "name": "color attachment"},
-            {"value": 32, "name": "depth stencil attachment"}
+            {"value": 16, "name": "output attachment"},
+            {"value": 32, "name": "input attachment"}
         ]
     },
     "texture view": {
--- a/next.json
+++ b/next.json
@@ -921,8 +921,7 @@
             {"value": 2, "name": "transfer dst"},
             {"value": 4, "name": "sampled"},
             {"value": 8, "name": "storage"},
-            {"value": 16, "name": "output attachment"},
-            {"value": 32, "name": "input attachment"}
+            {"value": 16, "name": "output attachment"}
         ]
     },
     "texture view": {
--- a/src/backend/CommandBufferStateTracker.cpp
+++ b/src/backend/CommandBufferStateTracker.cpp
@@ -148,15 +148,11 @@ namespace backend {
             }
 
             auto* texture = tv->GetTexture();
-            if (texture->HasFrozenUsage(nxt::TextureUsageBit::ColorAttachment)) {
-                continue;
-            }
-            if (!texture->IsTransitionPossible(nxt::TextureUsageBit::ColorAttachment)) {
-                builder->HandleError("Can't transition attachment to ColorAttachment usage");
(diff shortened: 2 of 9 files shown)


next.json
```diff
@@ -921,8 +921,8 @@
             {"value": 2, "name": "transfer dst"},
             {"value": 4, "name": "sampled"},
             {"value": 8, "name": "storage"},
-            {"value": 16, "name": "color attachment"},
-            {"value": 32, "name": "depth stencil attachment"}
+            {"value": 16, "name": "output attachment"},
+            {"value": 32, "name": "input attachment"}
```
Kangz: Maybe we could skip adding input attachments for now since we are not going to implement them for a while.

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -352,6 +354,18 @@ namespace backend {
         return true;
     }
 
+    bool CommandBufferStateTracker::EnsureTextureUsage(TextureBase* texture, nxt::TextureUsageBit usage) {
```
Kangz: nit: EnsureTextureUsage is weird. Maybe "PerformImplicitTextureUsageTransition"?

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -274,6 +270,10 @@ namespace backend {
     }
 
     bool CommandBufferStateTracker::SetBindGroup(uint32_t index, BindGroupBase* bindgroup) {
+        if (!HavePipeline()) {
```
Kangz: nit: I'm not sure we want to add this constraint. It doesn't help any backend (Metal and Vulkan don't care, D3D12 already needs to reset bindgroups on any layout change).

kainino0x:
I'm going to merge so I can move forward more easily. I'll rename EnsureTextureUsage in a followup if we decide on a different name.

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -406,17 +398,29 @@ namespace backend {
         if (aspects[VALIDATION_ASPECT_BIND_GROUPS]) {
             return true;
         }
-        if (bindgroupsSet.all()) {
-            aspects.set(VALIDATION_ASPECT_BIND_GROUPS);
-            return true;
+        // Assumes we have a pipeline already
+        if (!bindgroupsSet.all()) {
+            printf("%s:%d\n", __FUNCTION__, __LINE__);
```
Kangz: nit: temporary code?

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -406,17 +398,29 @@ namespace backend {
...
+            return false;
         }
-        return false;
+        for (size_t i = 0; i < bindgroups.size(); ++i) {
+            if (auto* bindgroup = bindgroups[i]) {
+                // TODO(kainino@chromium.org): bind group compatibility
+                if (bindgroup->GetLayout() != lastPipeline->GetLayout()->GetBindGroupLayout(i)) {
+                    printf("%s:%d: i=%zu\n", __FUNCTION__, __LINE__, i);
```
Kangz: same?

Kangz:
SGTM thanks

src/backend/CommandBufferStateTracker.cpp
kainino0x: oops. thanks.

PR #70, opened 2017-07-08
Introduce Begin/EndComputePass

kainino0x:
`SetPipeline(computePipeline)` is gated on `BeginComputePass`, so that should cover the necessary validation, at least for `Dispatch`.

### DIFF
--- a/src/backend/CommandBufferStateTracker.cpp
+++ b/src/backend/CommandBufferStateTracker.cpp
@@ -61,7 +61,7 @@ namespace backend {
 
     bool CommandBufferStateTracker::ValidateCanDispatch() {
         constexpr ValidationAspects requiredAspects =
-            1 << VALIDATION_ASPECT_COMPUTE_PIPELINE |
+            1 << VALIDATION_ASPECT_COMPUTE_PIPELINE | // implicitly requires COMPUTE_PASS
             1 << VALIDATION_ASPECT_BIND_GROUPS;
         if ((requiredAspects & ~aspects).none()) {
             // Fast return-true path if everything is good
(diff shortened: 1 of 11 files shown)


src/backend/metal/CommandBufferMTL.mm
```diff
@@ -141,6 +141,13 @@ void EndRenderPass() {
         uint32_t currentSubpass = 0;
         while (commands.NextCommandId(&type)) {
             switch (type) {
+                case Command::BeginComputePass:
+                    {
+                        commands.NextCommand<BeginComputePassCmd>();
+                        // XXX: begin compute pass
```
Kangz: Can you make it a TODO(kainino) ... here and below?

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -61,7 +61,7 @@ namespace backend {
 
     bool CommandBufferStateTracker::ValidateCanDispatch() {
```
Kangz: We should check that we aren't in a compute pass at the end of ValidateGetResult.

src/backend/metal/CommandBufferMTL.mm
kainino0x: Oh yeah. Actually, let me see if I can just DO this TODO.

src/backend/CommandBufferStateTracker.cpp
kainino0x: Done

kainino0x:
Done, could you TAL at the last commit?

PR #71, opened 2017-07-10
Add reference counting to DeviceBase

austinEng:
Adds reference counting to the device and deletes itself when it reaches 0
D3D12 backend waits for GPU execution to complete when destroyed

### DIFF
--- a/src/backend/Device.h
+++ b/src/backend/Device.h
@@ -27,7 +27,7 @@ namespace backend {
     class DeviceBase {
         public:
             DeviceBase();
-            ~DeviceBase();
+            virtual ~DeviceBase();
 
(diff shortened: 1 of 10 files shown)


src/backend/Device.h
```diff
@@ -52,6 +52,8 @@ namespace backend {
             virtual TextureViewBase* CreateTextureView(TextureViewBuilder* builder) = 0;
 
             virtual void TickImpl() = 0;
+            virtual void ReferenceImpl() = 0;
```
Kangz: nit: I don't think we'll ever need to do something inside these Impl() so we could remove them

Kangz:
LGTM!

PR #72, opened 2017-07-10
D3D12 Render Targets

austinEng:
Implements BeginRenderSubpass on the D3D12 backend. Descriptors for render target and depth stencil views are recorded in a descriptor heap for each framebuffer. For now, we still have the hack where no attachment renders to the backbuffer, so the CommandBuffer records those when necessary when it is submitted.

This PR also enables input states for D3D12 which are mostly working. One failure seems to be happening because our texture copies are not yet correct.

### DIFF
--- a/src/backend/d3d12/CommandBufferD3D12.cpp
+++ b/src/backend/d3d12/CommandBufferD3D12.cpp
@@ -192,7 +192,7 @@ namespace d3d12 {
                     D3D12_DESCRIPTOR_HEAP_TYPE_SAMPLER);
             }
         }
-    
+
         D3D12_TEXTURE_COPY_LOCATION D3D12PlacedTextureCopyLocation(BufferCopyLocation& bufferLocation, Texture* texture, const TextureCopyLocation& textureLocation) {
             D3D12_TEXTURE_COPY_LOCATION d3d12Location;
             d3d12Location.pResource = ToBackend(bufferLocation.buffer.Get())->GetD3D12Resource().Get();
@@ -252,6 +252,7 @@ namespace d3d12 {
 
         RenderPass* currentRenderPass = nullptr;
         Framebuffer* currentFramebuffer = nullptr;
+        uint32_t currentSubpass = 0;
 
         while(commands.NextCommandId(&type)) {
             switch (type) {
@@ -266,6 +267,7 @@ namespace d3d12 {
                         BeginRenderPassCmd* beginRenderPassCmd = commands.NextCommand<BeginRenderPassCmd>();
                         currentRenderPass = ToBackend(beginRenderPassCmd->renderPass.Get());
                         currentFramebuffer = ToBackend(beginRenderPassCmd->framebuffer.Get());
+                        currentSubpass = 0;
 
                         uint32_t width = currentFramebuffer->GetWidth();
                         uint32_t height = currentFramebuffer->GetHeight();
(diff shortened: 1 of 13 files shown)


src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -258,20 +259,74 @@ namespace d3d12 {
...
+                            ComPtr<ID3D12Resource> texture;
+                            if (auto textureView = currentFramebuffer->GetTextureView(attachment)) {
+                                texture = ToBackend(textureView->GetTexture())->GetD3D12Resource();
+                            } else {
+                                usingBackbuffer = true;
+                                texture = device->GetCurrentTexture();
+                            }
+
+                            D3D12_CPU_DESCRIPTOR_HANDLE rtvHandle = rtvHeap.GetCPUHandle(rtvIndex++);
+                            D3D12_RENDER_TARGET_VIEW_DESC rtvDesc;
```
Kangz: nit: this should be created from the view as it will eventually have information about the mipSlice and planeSlice. Something like a ```TextureView::GetRTVDesc```, no need to cache it in the texture view though.

src/backend/d3d12/D3D12Backend.cpp
```diff
@@ -155,14 +155,17 @@ namespace d3d12 {
...
-        return renderTargetDescriptor;
+    ComPtr<ID3D12Resource> Device::GetCurrentTexture() {
+        return nextTexture;
     }
 
-    void Device::SetNextRenderTargetDescriptor(D3D12_CPU_DESCRIPTOR_HANDLE renderTargetDescriptor) {
-        this->renderTargetDescriptor = renderTargetDescriptor;
-        static const float clearColor[] = { 0.0f, 0.0f, 0.0f, 1.0f };
-        GetPendingCommandList()->ClearRenderTargetView(renderTargetDescriptor, clearColor, 0, nullptr);
+    ComPtr<ID3D12Resource> Device::GetCurrentDSVTexture() {
```
Kangz: nit: DSV is DepthStencilView, here it is more ```GetCurrentDepthTexture```

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -19,14 +19,15 @@
...
-#include "backend/d3d12/DescriptorHeapAllocator.h"
+#include "backend/d3d12/FramebufferD3D12.h"
 #include "backend/d3d12/InputStateD3D12.h"
 #include "backend/d3d12/PipelineD3D12.h"
 #include "backend/d3d12/PipelineLayoutD3D12.h"
 #include "backend/d3d12/SamplerD3D12.h"
 #include "backend/d3d12/TextureD3D12.h"
 
-#include "ResourceAllocator.h"
+#include "backend/d3d12/DescriptorHeapAllocator.h"
```
Kangz: nit: these headers can be ordered with the rest of the d3d12/ include above.

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -258,20 +259,74 @@ namespace d3d12 {
...
                         commandList->RSSetViewports(1, &viewport);
                         commandList->RSSetScissorRects(1, &scissorRect);
-                        commandList->OMSetRenderTargets(1, &device->GetCurrentRenderTargetDescriptor(), FALSE, nullptr);
                     }
                     break;
 
                 case Command::BeginRenderSubpass:
                     {
                         commands.NextCommand<BeginRenderSubpassCmd>();
+                        const auto& subpassInfo = currentRenderPass->GetSubpassInfo(currentSubpass);
```
Kangz: Just saw that you did part of that in the next commit, though more things could be moved in d3d12::Framebuffer

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -275,57 +276,41 @@ namespace d3d12 {
...
                             uint32_t attachment = subpassInfo.colorAttachments[index];
                             const auto& attachmentInfo = currentRenderPass->GetAttachmentInfo(attachment);
 
-                            ComPtr<ID3D12Resource> texture;
-                            if (auto textureView = currentFramebuffer->GetTextureView(attachment)) {
-                                texture = ToBackend(textureView->GetTexture())->GetD3D12Resource();
-                            } else {
+                            D3D12_CPU_DESCRIPTOR_HANDLE rtvHandle = currentFramebuffer->GetRTVHandle(attachment);
+                            rtvHandles[rtvCount++] = rtvHandle;
+                            if (!currentFramebuffer->GetTextureView(attachment)) {
```
Kangz: This handling of the backbuffer could be hidden in Framebuffer so that the code here is almost the final version (except the clear hack).

src/backend/d3d12/FramebufferD3D12.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+                device->GetD3D12Device()->CreateRenderTargetView(device->GetCurrentTexture().Get(), &rtvDesc, rtvHandle);
+            }
+        }
+
+        if (usingBackbuffer) {
+            args.dsv = dsvHeap.GetCPUHandle(0);
+        }
+
+        return std::move(args);
```
Kangz: no need to std::move because the calling convention makes it so in the assembly, the return value is allocated on the parent's stack frame and a pointer to it is passed here.

src/backend/d3d12/FramebufferD3D12.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+        RenderPass* renderPass = ToBackend(GetRenderPass());
+        uint32_t attachmentCount = renderPass->GetAttachmentCount();
+        rtvHeap = device->GetDescriptorHeapAllocator()->AllocateCPUHeap(D3D12_DESCRIPTOR_HEAP_TYPE_RTV, attachmentCount);
+
+        bool usingBackbuffer = false; // HACK(enga@google.com): workaround for not having depth attachments
+        uint32_t rtvIndex = 0;
+
+        for (uint32_t attachment = 0; attachment < attachmentCount; ++attachment) {
+            if (auto textureView = GetTextureView(attachment)) {
```
Kangz: if (auto...) is a bit non-standard.

kainino0x: This was copied from my code. This code goes away when we add WSI, anyway.

PR #73, opened 2017-07-11
OpenGL backend: implement render passes

kainino0x:
Probably not the most efficient implementation, but works.

Issues:
* [ ] Doesn't seem to render until the second frame. No clue why, yet.
* [ ] Hardcoded 640x480 in more places.
* [ ] Creates new framebuffer objects for every subpass every frame. This should really be done at command buffer build time. (Is that currently possible?)

### DIFF
--- a/src/backend/opengl/CommandBufferGL.cpp
+++ b/src/backend/opengl/CommandBufferGL.cpp
@@ -88,6 +88,10 @@ namespace opengl {
                 case Command::BeginRenderSubpass:
                     {
                         commands.NextCommand<BeginRenderSubpassCmd>();
+                        // TODO(kainino@chromium.org): possible future
+                        // optimization: create these framebuffers at command
+                        // buffer build time so they don't have to be created
+                        // and destroyed at draw time.
                         glGenFramebuffers(1, &currentFBO);
                         glBindFramebuffer(GL_DRAW_FRAMEBUFFER, currentFBO);
 
--- a/src/backend/opengl/CommandBufferGL.cpp
+++ b/src/backend/opengl/CommandBufferGL.cpp
@@ -89,9 +89,10 @@ namespace opengl {
                     {
                         commands.NextCommand<BeginRenderSubpassCmd>();
                         // TODO(kainino@chromium.org): possible future
-                        // optimization: create these framebuffers at command
-                        // buffer build time so they don't have to be created
-                        // and destroyed at draw time.
+                        // optimization: create these framebuffers at
+                        // Framebuffer build time (or maybe CommandBuffer build
+                        // time) so they don't have to be created and destroyed
+                        // at draw time.
                         glGenFramebuffers(1, &currentFBO);
                         glBindFramebuffer(GL_DRAW_FRAMEBUFFER, currentFBO);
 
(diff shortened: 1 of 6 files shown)


src/backend/opengl/OpenGLBackend.cpp
```diff
@@ -100,6 +108,46 @@ namespace opengl {
...
+        glClearColor(0, 0, 0, 1);
+        glStencilMask(0xff);
+        glClearStencil(0);
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
+    }
+
+    void Device::InitBackbuffer() {
+        glGenTextures(1, &backTexture);
+        glBindTexture(GL_TEXTURE_2D, backTexture);
+        glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB8, 640, 480, 0, GL_RGB, GL_UNSIGNED_BYTE, nullptr);
```
Kangz: The backbuffer could have an alpha channel too :)

src/utils/OpenGLBinding.cpp
```diff
@@ -42,11 +44,18 @@ namespace utils {
...
                 backend::opengl::Init(reinterpret_cast<void*(*)(const char*)>(glfwGetProcAddress), procs, device);
+
+                backendDevice = *device;
+                backend::opengl::InitBackbuffer(backendDevice);
             }
             void SwapBuffers() override {
+                backend::opengl::CommitBackbuffer(backendDevice);
                 glfwSwapBuffers(window);
-                backend::opengl::HACKCLEAR();
+                backend::opengl::HACKCLEAR(backendDevice);
```
Kangz: Maybe we should HACKCLEAR just after InitBackbuffer too?

Kangz:
I don't think creating framebuffers all the time is a big problem, they are just container objects in GL. What we could do if we really wanted to reuse framebuffers, would be to create one framebuffer per subpass on nxtFramebuffer creation.

src/utils/OpenGLBinding.cpp
kainino0x: Ah right, forgot to do this after I moved HACKCLEAR into the device. Unfortunately it doesn't seem to fix the first frame(s) being black.

Kangz:
It depends on how renderpass compatibility is defined, I don't think it should interact a lot with how the framebuffer is set up, it is mostly about load and store ops IIRC.

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -88,6 +88,10 @@ namespace opengl {
                 case Command::BeginRenderSubpass:
                     {
                         commands.NextCommand<BeginRenderSubpassCmd>();
+                        // TODO(kainino@chromium.org): possible future
+                        // optimization: create these framebuffers at command
```
Kangz: see comment on PR.

kainino0x: ok. I'll edit this comment.

kainino0x:
Skipped CI because I only edited a comment.

PR #74, opened 2017-07-11
Windows Warnings

Kangz:
PTAL this fixes all warnings on Windows and does some improvements to the CMake files to work better with visual studio.

### DIFF
--- a/src/common/Assert.h
+++ b/src/common/Assert.h
@@ -35,7 +35,7 @@ void HandleAssertionFailure(const char* file, const char* function, int line, co
 
 // MSVC triggers a warning in /W4 for do {} while(0). SDL worked around this by using
 // // (0,0) and points out that it looks like an owl face.
-#if defined(_MSC_VER)
+#if defined(NXT_COMPILER_MSVC)
     #define NXT_ASSERT_LOOP_CONDITION (0,0)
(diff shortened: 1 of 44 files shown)


src/common/Assert.h
```diff
@@ -66,9 +66,11 @@ void HandleAssertionFailure(const char* file, const char* function, int line, co
 #endif
 
 #define NXT_ASSERT(condition) NXT_ASSERT_CALLSITE_HELPER(__FILE__, __func__, __LINE__, condition)
+#define NXT_UNREACHABLE() NXT_ASSERT(false && "Unreachable code hit"); NXT_BUILTIN_UNREACHABLE()
```
kainino0x: Semicolon in a macro is a bit scary. Is it possible to have
```
#define NXT_UNREACHABLE() do { NXT_ASSERT(false && "Unreachable code hit"); NXT_BUILTIN_UNREACHABLE(); } while(0)
```
or does that break `__builtin_unreachable`?

Kangz: Done, it didn't change the behavior of __builtin_unreachable()

Kangz:
Good point about ```-Wconversion``` I can look at adding that next.

PR #76, opened 2017-07-12
Input state fix

Kangz:


src/backend/Framebuffer.cpp
```diff
@@ -75,6 +75,14 @@ namespace backend {
                     return nullptr;
                 }
                 usingBackbufferHack = true;
+                continue;
+            }
+
+            // Adjust for the mip-level onces that is supported.
```
kainino0x: Is this a TODO?

Kangz: Yes, done.

Kangz:
Thanks!

PR #77, opened 2017-07-12
Add depth-stencil textures and attachments

kainino0x:
Related: #29 

* [x] Modify D3D12 backend as necessary once #72 lands

### DIFF
--- a/next.json
+++ b/next.json
@@ -591,7 +591,7 @@
                 "name": "set attachment",
                 "args": [
                     {"name": "attachment slot", "type": "uint32_t"},
-                    {"name": "texture views", "type": "texture view"}
+                    {"name": "texture view", "type": "texture view"}
                 ]
             }
         ]
@@ -760,6 +760,13 @@
                 "args": [
                     {"name": "subpass index", "type": "uint32_t"},
                     {"name": "output attachment location", "type": "uint32_t"},
+                    {"name": "attachment slot", "type": "uint32_t"}
+                ]
+            },
+            {
+                "name": "subpass set depth stencil attachment",
+                "args": [
+                    {"name": "subpass index", "type": "uint32_t"},
                     {"name": "attachment slot", "type": "uint32_t"}
                 ]
             }
@@ -936,7 +943,8 @@
     "texture format": {
         "category": "enum",
         "values": [
-            {"value": 0, "name": "r8 g8 b8 a8 unorm"}
+            {"value": 0, "name": "r8 g8 b8 a8 unorm"},
+            {"value": 1, "name": "d32 float s8 uint"}
         ]
     },
     "vertex format": {
--- a/src/backend/RenderPass.cpp
+++ b/src/backend/RenderPass.cpp
@@ -147,4 +147,26 @@ namespace backend {
         subpasses[subpass].colorAttachments[outputAttachmentLocation] = attachmentSlot;
     }
 
+    void RenderPassBuilder::SubpassSetDepthStencilAttachment(uint32_t subpass, uint32_t attachmentSlot) {
+        if ((propertiesSet & RENDERPASS_PROPERTY_SUBPASS_COUNT) == 0) {
+            HandleError("Render pass subpass count not set yet");
+            return;
+        }
+        if (subpass >= subpasses.size()) {
+            HandleError("Subpass index out of bounds");
+            return;
+        }
+        if (attachmentSlot >= attachments.size()) {
+            HandleError("Subpass attachment slot out of bounds");
+            return;
+        }
+        if (subpasses[subpass].depthStencilAttachmentSet) {
+            HandleError("Subpass depth-stencil attachment already set");
+            return;
+        }
+
+        subpasses[subpass].depthStencilAttachmentSet = true;
+        subpasses[subpass].depthStencilAttachment = attachmentSlot;
+    }
+
 }
--- a/src/backend/RenderPass.cpp
+++ b/src/backend/RenderPass.cpp
@@ -126,6 +126,10 @@ namespace backend {
             HandleError("Render pass subpass count not set yet");
             return;
         }
+        if ((propertiesSet & RENDERPASS_PROPERTY_ATTACHMENT_COUNT) == 0) {
+            HandleError("Render pass attachment count not set yet");
+            return;
+        }
         if (subpass >= subpasses.size()) {
             HandleError("Subpass index out of bounds");
             return;
@@ -150,6 +154,10 @@ namespace backend {
     void RenderPassBuilder::SubpassSetDepthStencilAttachment(uint32_t subpass, uint32_t attachmentSlot) {
         if ((propertiesSet & RENDERPASS_PROPERTY_SUBPASS_COUNT) == 0) {
             HandleError("Render pass subpass count not set yet");
+            return;
+        }
+        if ((propertiesSet & RENDERPASS_PROPERTY_ATTACHMENT_COUNT) == 0) {
+            HandleError("Render pass attachment count not set yet");
             return;
         }
         if (subpass >= subpasses.size()) {
--- a/src/backend/Texture.h
+++ b/src/backend/Texture.h
@@ -24,6 +24,8 @@
 namespace backend {
 
     size_t TextureFormatPixelSize(nxt::TextureFormat format);
+    bool TextureFormatHasDepth(nxt::TextureFormat format);
+    bool TextureFormatHasStencil(nxt::TextureFormat format);
 
     class TextureBase : public RefCounted {
         public:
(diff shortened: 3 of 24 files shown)


kainino0x:
Note that any application which uses more than one renderpass per frame will get its depth/stencil buffers cleared at every renderpass. This only manifests in glTFViewer, in which parts of the model will be drawn in an arbitrary order, with correct depth testing per-part, but incorrect depth testing between parts.

src/backend/RenderPass.cpp
```diff
@@ -147,4 +147,26 @@ namespace backend {
...
+    void RenderPassBuilder::SubpassSetDepthStencilAttachment(uint32_t subpass, uint32_t attachmentSlot) {
+        if ((propertiesSet & RENDERPASS_PROPERTY_SUBPASS_COUNT) == 0) {
+            HandleError("Render pass subpass count not set yet");
+            return;
+        }
+        if (subpass >= subpasses.size()) {
+            HandleError("Subpass index out of bounds");
+            return;
+        }
+        if (attachmentSlot >= attachments.size()) {
```
Kangz: We should check that the attachment count has been set before too.

next.json
```diff
@@ -762,6 +762,13 @@
                     {"name": "output attachment location", "type": "uint32_t"},
                     {"name": "attachment slot", "type": "uint32_t"}
                 ]
+            },
+            {
+                "name": "subpass set depth stencil attachment",
```
Kangz: Any idea how to handle separate depth and stencil attachments vs. combined?

src/backend/Texture.h
```diff
@@ -44,6 +44,7 @@ namespace backend {
             void UpdateUsageInternal(nxt::TextureUsageBit usage);
 
             static bool IsDepthFormat(nxt::TextureFormat format);
+            static bool IsStencilFormat(nxt::TextureFormat format);
```
Kangz: nit: Can you move both of these to be outside of the TextureBase class and next to TextureFormatPixelSize?

next.json
kainino0x: Oh yeah, I intended to split this. From the APIs I've looked at, you just set the two separately, and they can point at the same TextureView if it's a combined format.

Kangz: As we talked about in Hangouts, D3D12 only allows one resource for depth and/or stencil, so we can go with a single entry point.

kainino0x:
Wrote code for D3D12 backend... but on my mac, so I haven't compiled it. If I'm lucky, I can make it build just by watching appveyor, but I'll need someone to run it for me at some point.

kainino0x:
jk, turns out the D3D12 backend doesn't compile on appveyor (I didn't know).

@austinEng when you get a chance, could you pull from my branch and build it? If you have time, could you get it working and maybe make a PR into my depthstencil branch?

Kangz:
Erhm, AppVeyor should compile the D3D12 backend, I must have messed something up with the feature defines :/

src/backend/d3d12/FramebufferD3D12.cpp
```diff
@@ -41,31 +40,26 @@ namespace d3d12 {
...
             } else {
                 // TODO(enga@google.com) no attachment. This will use the backbuffer. Remove when this hack is removed
-                usingBackbuffer = true;
                 rtvIndex++;
             }
         }
 
-        // TODO(enga@google.com): load depth attachment from subpass
-        if (usingBackbuffer) {
+        if (info.depthStencilAttachmentSet) {
```
austinEng: There is no `SubpassInfo& info` defined here.

Unless I am mistaken, `renderPass->GetAttachmentCount()`, is the total number of combined color and depth/stencil attachments. We need to separately count the number of color and the number of depth/stencil attachments and then allocate RTV and DSV descriptor heaps of the appropriate sizes.

src/backend/d3d12/TextureD3D12.cpp
```diff
@@ -164,8 +170,19 @@ namespace d3d12 {
...
+        rtvDesc.Flags = D3D12_RTV_FLAG_NONE;
         return rtvDesc;
     }
 
+    D3D12_DEPTH_STENCIL_VIEW_DESC TextureView::GetDSVDescriptor() {
+        D3D12_DEPTH_STENCIL_VIEW_DESC dsvDesc;
+        dsvDesc.Format = ToBackend(GetTexture())->GetD3D12Format();
+        dsvDesc.ViewDimension = D3D12_DSV_DIMENSION_TEXTURE2D;
+        dsvDesc.Texture2D.MipSlice = 0;
+        dsvDesc.Texture2D.PlaneSlice = 0;
```
austinEng: D3D12_TEX2D_DSV does not have a PlaneSlice

src/backend/d3d12/TextureD3D12.cpp
```diff
@@ -164,8 +170,19 @@ namespace d3d12 {
         rtvDesc.ViewDimension = D3D12_RTV_DIMENSION_TEXTURE2D;
         rtvDesc.Texture2D.MipSlice = 0;
         rtvDesc.Texture2D.PlaneSlice = 0;
+        rtvDesc.Flags = D3D12_RTV_FLAG_NONE;
```
austinEng: D3D12_RENDER_TARGET_VIEW_DESC  does not have Flags

src/backend/d3d12/FramebufferD3D12.cpp
```diff
@@ -41,31 +40,26 @@ namespace d3d12 {
...
 
-            D3D12_DEPTH_STENCIL_VIEW_DESC dsvDesc;
-            dsvDesc.Format = DXGI_FORMAT_D32_FLOAT_S8X24_UINT;
-            dsvDesc.ViewDimension = D3D12_DSV_DIMENSION_TEXTURE2D;
-            dsvDesc.Texture2D.MipSlice = 0;
-            dsvDesc.Flags = D3D12_DSV_FLAG_NONE;
+            uint32_t attachment = info.depthAttachment;
+            auto textureView = currentFramebuffer->GetTextureView(attachment);
+            ComPtr<ID3D12Resource> texture = ToBackend(textureView->GetTexture())->GetD3D12Resource();
+            D3D12_CPU_DESCRIPTOR_HANDLE dsvHandle = dsvHeap.GetCPUHandle(0);
```
austinEng: And then this will need to be something like `dsvHeap.GetCPUHandle(depthStencilAttachmentIndex);`

src/backend/d3d12/FramebufferD3D12.cpp
```diff
@@ -87,8 +79,7 @@ namespace d3d12 {
                 device->GetD3D12Device()->CreateRenderTargetView(device->GetCurrentTexture().Get(), &rtvDesc, rtvHandle);
             }
         }
-
-        if (usingBackbuffer) {
+        if (subpassInfo.depthStencilAttachmentSet) {
             args.dsv = dsvHeap.GetCPUHandle(0);
```
austinEng: And then we'll need to remember the depthStencilIndex for each subpass so we can get the correct handle here.

src/backend/d3d12/FramebufferD3D12.cpp
kainino0x: oops, you're right, this is totally wrong. Thanks for catching it. I'll do as you suggested.

src/backend/d3d12/TextureD3D12.cpp
kainino0x: done

src/backend/d3d12/FramebufferD3D12.cpp
kainino0x: got it.

src/backend/d3d12/TextureD3D12.cpp
kainino0x: done

src/backend/d3d12/FramebufferD3D12.cpp
kainino0x: ah, this is somewhat annoying. We have an index in subpassInfo but it's for all attachments, so it won't work for dsvHeap. I'll think about this.

src/backend/d3d12/FramebufferD3D12.h
```diff
@@ -41,6 +43,9 @@ namespace d3d12 {
             Device* device;
             DescriptorHeapHandle rtvHeap;
```
Kangz: nit: these should be null initialized so that they are nullptr if there are no color / DS attachments, which would make crashes easier to understand.

kainino0x: `DescriptorHeapHandle* rtvHeap = nullptr;` or `DescriptorHeapHandle rtvHeap = {}`?

Actually I'm confused about DescriptorHeapHandle. Where is this `device` coming from??
https://github.com/google/nxt-standalone/blob/master/src/backend/d3d12/DescriptorHeapAllocator.cpp#L27
and also what is the default constructor useful for?

kainino0x: Fixed (#82)

PR #79, opened 2017-07-14
Fix CMake OS dependent default options.

Kangz:
The defaults were setting local variables, while the "option" command
uses the values from the CMake cache. This made the CI not build the
D3D12 or Metal backend on their platforms.


Kangz:
__assume(0) is a hint to the compiler, but there should be debugbreak() that "crashes" in Debug.

AFAIK there is no way to change Debug vs. Release from inside visual studio when using CMake :/

kainino0x:
CMAKE_BUILD_TYPE isn't used for multi-configuration builders like Visual Studio. Took some searching around, but the way to do this is:
https://stackoverflow.com/a/24892791
https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html

PR #85, opened 2017-07-18
Add T2B and B2T end2end tests

austinEng:


src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+            CheckRegion(0, 0);
+            CheckRegion(0, height - regionHeight);
+            CheckRegion(width - regionWidth, 0);
+            CheckRegion(width - regionWidth, height - regionHeight);
+            CheckRegion((width - regionWidth) / 3, (height - regionHeight) / 7);
+            CheckRegion((width - regionWidth) / 7, (height - regionHeight) / 3);
+        }
+
+        void TestMipLevel(uint32_t width, uint32_t height, uint32_t level) {
```
kainino0x: nit: maybe comment this?

src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,418 @@
...
+};
+
+class CopyTests_T2B : public CopyTests {
+    protected:
+
+        // Test that on initialization, the texture data is as expected
+        void TestFullTexture(uint32_t width, uint32_t height) {
+            SourceTexture source = CreateSourceTexture(width, height);
+            std::vector<RGBA8> expected = PackedRegion(source.data.data(), width, 0, 0, width, height);
+            EXPECT_TEXTURE_RGBA8_EQ(expected.data(), source.texture, 0, 0, width, height, 0) << "Failed initialization of " << width << "x" << height << " texture" << std::endl;
```
Kangz: It looks like a lot of the test relies on the internal implementation of EXPECT_TEXTURE_RGBA to do a texture to buffer copy with specific parameters. Can we remove this assumption and do the copy here directly?

src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,418 @@
...
+#include <array>
+#include "common/Constants.h"
+#include "common/Math.h"
+#include "utils/NXTHelpers.h"
+
+class CopyTests : public NXTTest {
+    protected:
+        static constexpr unsigned int kBytesPerTexel = 4;
+
+        struct SourceTexture {
```
Kangz: nit: SourceTexture could be renamed to account for B->T copies done in this test.

src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,418 @@
...
+        void TestFullTexture(uint32_t width, uint32_t height) {
+            SourceTexture source = CreateSourceTexture(width, height);
+            std::vector<RGBA8> expected = PackedRegion(source.data.data(), width, 0, 0, width, height);
+            EXPECT_TEXTURE_RGBA8_EQ(expected.data(), source.texture, 0, 0, width, height, 0) << "Failed initialization of " << width << "x" << height << " texture" << std::endl;
+        }
+
+        // Inside a width x height texture, check the following pixels:
+        // top left corner, top right corner, bottom left corner, bottom right corner
+        // and arbitrarily 1/3 across 1/7 down and 1/7 across and 1/3 down
+        void TestPixels(uint32_t width, uint32_t height) {
```
Kangz: This and TestFullTexture are basically two instances of TestTextureRegion, overall I feel we could factor a lot of things between the different tests, and almost have a single function "DoTest(TestParams)" with the test params having the buffer offset, pitch, texture size, texture level and copy region.

src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,418 @@
...
+                .GetResult();
+
+            queue.Submit(1, &commands);
+
+            EXPECT_TEXTURE_RGBA8_EQ(expected.data(), texture, 0, 0, width, height, 0) << "Failed to copy to " << width << "x" << height << " texture with buffer row pitch " << bufferRowPitch << std::endl;
+        }
+};
+
+// Test that creating and reading back an entire texture with 256-byte aligned dimensions works
+TEST_P(CopyTests_T2B, FullTextureAligned) {
```
Kangz: T2B versions of the test are testing the texture-related parameters and B2T are testing the buffer-related ones. We basically have half-coverage of the T<->B copies. Can we make both T->B and B->T test both?

src/tests/NXTTest.cpp
```diff
@@ -193,7 +195,8 @@ void NXTTest::AddTextureExpectation(const char* file, int line, const nxt::Textu
     deferred.rowPitch = rowPitch;
     deferred.expectation = expectation;
 
-    deferredExpectations.push_back(deferred);
+    deferredExpectations.push_back(std::move(deferred));
```
austinEng: These look the same to me

Kangz: They do now, that was when there was an emplace_back on one side and a push_back on the other.

src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,502 @@
...
+            for (unsigned int y = 0; y < height; ++y) {
+                for (unsigned int x = 0; x < width; ++x) {
+                    unsigned int i = x + y * texelsPerRow;
+                    data[i] = RGBA8(x % 255, y % 255, 0, 255);
+                }
+            }
+        }
+
+        void DoTest(const TextureSpec& textureSpec, const BufferSpec& bufferSpec) {
+            nxt::Texture texture = device.CreateTextureBuilder()
```
Kangz: nit: would love some comments to highlight what the different part of this function do (same for B2T). Also exaplin that the upload buffer is for the given mip level.

src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+
+};
+
+class CopyTests_B2T : public CopyTests {
+protected:
+    static void FillBufferData(RGBA8* data, size_t count) {
+        for (size_t i = 0; i < count; ++i) {
+            data[i] = RGBA8(
+                static_cast<uint8_t>(i % 255),
```
Kangz: Did you mean to do all of that with 256 instead? Here the 256th element will have (1, 1, 0, 0) wich looks like 257.

src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,502 @@
...
+        }
+};
+
+class CopyTests_T2B : public CopyTests {
+    protected:
+        static void FillTextureData(uint32_t width, uint32_t height, uint32_t texelsPerRow, RGBA8* data) {
+            for (unsigned int y = 0; y < height; ++y) {
+                for (unsigned int x = 0; x < width; ++x) {
+                    unsigned int i = x + y * texelsPerRow;
+                    data[i] = RGBA8(x % 255, y % 255, 0, 255);
```
Kangz: Can it be RGBA8(x%256, y%256, x/256, y/256) so that we are sure each pixel is unique?

src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,502 @@
...
+
+// Test that copying without a 512-byte aligned buffer offset works
+TEST_P(CopyTests_T2B, OffsetBufferUnaligned) {
+    if (IsD3D12()) {
+        printf("TODO(enga@google.com): Test skipped because unaligned buffer offsets not supported in D3D12\n");
+        return;
+    }
+    constexpr uint32_t kWidth = 128;
+    constexpr uint32_t kHeight = 128;
+    for (uint32_t i = kBytesPerTexel; i < 512; i += kBytesPerTexel * 3) {
```
Kangz: That looks like a lot of iterations if it takes much longer than the other tests, can we have the increment be bigger?

src/tests/end2end/CopyTests.cpp
```diff
@@ -0,0 +1,517 @@
...
+};
+
+class CopyTests_B2T : public CopyTests {
+protected:
+    static void FillBufferData(RGBA8* data, size_t count) {
+        for (size_t i = 0; i < count; ++i) {
+            data[i] = RGBA8(
+                static_cast<uint8_t>(i % 256),
+                static_cast<uint8_t>((i / 256) % 256),
+                static_cast<uint8_t>((i / 256 / 256) % 255),
```
Kangz: nit: stray 255 :)

PR #86, opened 2017-07-18
Add Buffer offset validation for for T2B and B2T copies

austinEng:


src/backend/CommandBuffer.cpp
```diff
@@ -70,6 +70,16 @@ namespace backend {
             return true;
         }
 
+        bool ValidateBufferOffset(CommandBufferBuilder* builder, TextureBase* texture, const BufferCopyLocation& location) {
+            uint32_t texelSize = static_cast<uint32_t>(TextureFormatPixelSize(texture->GetFormat()));
```
Kangz: I wonder how this is going to work for RGB8Unorm, probably texel size will still be 4 in that case?

src/backend/CommandBuffer.cpp
```diff
@@ -437,6 +447,7 @@ namespace backend {
                             !ComputeTextureCopyBufferSize(this, copy->destination, copy->rowPitch, &bufferCopySize) ||
                             !ValidateCopyLocationFitsInTexture(this, copy->destination) ||
                             !ValidateCopySizeFitsInBuffer(this, copy->source, bufferCopySize) ||
+                            !ValidateBufferOffset(this, copy->destination.texture.Get(), copy->source) ||
```
Kangz: nit: ValidateTexelBufferOffset

src/tests/unittests/validation/CopyCommandsValidationTests.cpp
```diff
@@ -314,6 +314,30 @@ TEST_F(CopyCommandTest_B2T, IncorrectRowPitch) {
...
 }
 
+// Test B2T copies with incorrect buffer offset usage
+TEST_F(CopyCommandTest_B2T, IncorrectBufferOffset) {
+    uint32_t bufferSize = BufferSizeForTextureCopy(4, 4, 1);
+    nxt::Buffer source = CreateFrozenBuffer(bufferSize, nxt::BufferUsageBit::TransferSrc);
+    nxt::Texture destination = CreateFrozen2DTexture(16, 16, 5, nxt::TextureFormat::R8G8B8A8Unorm,
+                                                     nxt::TextureUsageBit::TransferDst);
+
+    {
```
Kangz: Can we have a control case with - 4 that Assert that it succeeds?

austinEng: CopyCommandTest_B2T, Success checks that

src/backend/CommandBuffer.cpp
austinEng: Done

src/backend/CommandBuffer.cpp
austinEng: Not sure what you mean. Why wouldn't it be?

src/tests/unittests/validation/CopyCommandsValidationTests.cpp
Kangz: It is ok to have redundant control cases if it helps readability by having locally an example of what works.

austinEng: Done

src/backend/CommandBuffer.cpp
Kangz: An RGB8Unorm format would only has 3 bytes of data, but we don't want to ask to align the buffer to 3. I assume the backing APIs will actually enforce an alignment of 4 bytes, but we'll have to investigate more eventually.

Kangz:
LGTM too

PR #87, opened 2017-07-19
Swap chains, part 1

kainino0x:
Adds the swap chain interfaces to the API without changing the behavior of anything else. This includes the C APIs for applications to provide swap chain implementations. Also adds stub implementations on every backend.

### DIFF
--- a/src/backend/SwapChain.cpp
+++ b/src/backend/SwapChain.cpp
@@ -1,0 +1,98 @@
+// Copyright 2017 The NXT Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "backend/SwapChain.h"
+
+#include "backend/Device.h"
+#include "backend/Texture.h"
+
+namespace backend {
+
+    // SwapChain
+
+    SwapChainBase::SwapChainBase(SwapChainBuilder* builder)
+            : device(builder->device), implementation(builder->implementation) {
+    }
+
+    SwapChainBase::~SwapChainBase() {
+    }
+
+    DeviceBase* SwapChainBase::GetDevice() {
+        return device;
+    }
+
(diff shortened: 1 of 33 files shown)


src/backend/SwapChain.cpp
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    void SwapChainBase::Configure(nxt::TextureFormat format, uint32_t width, uint32_t height) {
+        this->format = format;
+        this->width = width;
+        this->height = height;
+    }
+
+    TextureBase* SwapChainBase::GetNextTexture() {
+        auto* builder = device->CreateTextureBuilder();
```
Kangz: We should return an error if Configure wasn't called before.

src/backend/SwapChain.cpp
```diff
@@ -0,0 +1,98 @@
...
+        lastNextTexture = texture;
+        return texture;
+    }
+
+    void SwapChainBase::Present(TextureBase* texture) {
+        if (texture != lastNextTexture) {
+            device->HandleError("Tried to present something other than the last NextTexture");
+            return;
+        }
+
```
Kangz: We should check the texture is in the present usage.

src/backend/metal/ShaderModuleMTL.mm
```diff
@@ -29,6 +29,7 @@
         ExtractSpirvInfo(*compiler);
 
         std::string msl = compiler->compile();
+        printf("%s\n", msl.c_str());
```
Kangz: nit: stray printf

src/backend/metal/TextureMTL.mm
```diff
@@ -73,7 +73,12 @@ MTLTextureType MetalTextureType(nxt::TextureDimension dimension) {
         mtlTexture = [mtlDevice newTextureWithDescriptor:desc];
     }
 
+    Texture::Texture(TextureBuilder* builder, id<MTLTexture> mtlTexture)
+        : TextureBase(builder), mtlTexture(mtlTexture) {
+    }
+
     Texture::~Texture() {
+        // TODO(kainino@chromium.org): Maybe don't release when using the native texture constructor?
```
Kangz: Or we could add a ref in the WSI constructor, maybe.

src/backend/opengl/TextureGL.cpp
```diff
@@ -48,8 +48,18 @@ namespace opengl {
 
     // Texture
 
+    static GLuint GenTexture() {
```
Kangz: nit: this can be in the anonymous namespace above.

src/backend/metal/TextureMTL.mm
kainino0x: Oh, good idea. Works for this case, but what about the D3D12/GL backends?

PR #88, opened 2017-07-19
Correctly configure NXT_ENABLE_ASSERTS on multiconfiguration generators

kainino0x:
And also set a default build type for single-configuration generators.

We now set NXT_ENABLE_ASSERTS only on Debug builds or if NXT_ALWAYS_ASSERT is set.

https://github.com/google/nxt-standalone/pull/79#issuecomment-315430600


CMakeLists.txt
```diff
@@ -15,6 +15,11 @@
 cmake_minimum_required(VERSION 2.8)
 project(nxt C CXX)
 
+if(NOT CMAKE_BUILD_TYPE)
```
Kangz: nit: comment about the fact that by default the cmake build type can be ""

examples/CMakeLists.txt
```diff
@@ -24,7 +24,7 @@ function(add_nxt_sample target sources)
     target_link_libraries(${target} sample_utils)
     target_include_directories(${target} SYSTEM PRIVATE ${GLM_INCLUDE_DIR})
     target_include_directories(${target} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
-    NXTInternaltarget("examples" ${target})
+    NXTInternalTarget("examples" ${target})
```
Kangz: nit: eol change?

kainino0x: case change (but it's case insensitive so it doesn't actually matter)

PR #90, opened 2017-07-20
D3D12: Texture Copy Splitting for non 512-byte aligned copies

austinEng:


src/tests/unittests/d3d12/CopySplit.cpp
```diff
@@ -0,0 +1,375 @@
...
+                bool overlapY = RangesOverlap(a.texture.y, a.texture.y + a.texture.height, b.texture.y, b.texture.y + b.texture.height);
+                bool overlapZ = RangesOverlap(a.texture.z, a.texture.z + a.texture.depth, b.texture.z, b.texture.z + b.texture.depth);
+                ASSERT_TRUE(!overlapX || !overlapY || !overlapZ);
+            }
+        }
+    }
+
+    // Check that the union of the copy regions exactly covers the texture region
+    void ValidateTextureCoverage(const TextureSpec& textureSpec, const CopySplit& copySplit) {
+        ASSERT_TRUE(copySplit.count > 0);
```
Kangz: This doesn't check that all pixels are covered, just that they are all in bounds. Can we have two checks that validate that both the regions are in bounds of the texture, and another one that checks the number of pixel copied matches what we expect?

src/tests/unittests/d3d12/CopySplit.cpp
```diff
@@ -0,0 +1,375 @@
...
+        ASSERT_EQ(minX, textureSpec.x);
+        ASSERT_EQ(minY, textureSpec.y);
+        ASSERT_EQ(minZ, textureSpec.z);
+        ASSERT_EQ(maxX, textureSpec.x + textureSpec.width);
+        ASSERT_EQ(maxY, textureSpec.y + textureSpec.height);
+        ASSERT_EQ(maxZ, textureSpec.z + textureSpec.depth);
+    }
+
+    // Check that copies start in the buffer exactly at the buffer offset
+    void ValidateBufferOffset(const TextureSpec& textureSpec, const BufferSpec& bufferSpec, const CopySplit& copySplit) {
```
Kangz: This could be changed to check that for each copy the first pixel is correctly copied. This plus the no-overlap plus the everything in the region check, plus the check for the number of pixels, assure that the copies are correctly putting all the pixels at the correct places.

src/tests/unittests/d3d12/CopySplit.cpp
```diff
@@ -0,0 +1,375 @@
...
+                buffer.x * textureSpec.texelSize +
+                buffer.y * bufferSpec.rowPitch +
+                buffer.z * bufferSpec.rowPitch * textureSpec.height);
+        }
+
+        ASSERT_EQ(copySplit.offset + offset, bufferSpec.offset);
+    }
+   
+    // Validate there are exactly two copies if the copy straddles the row pitch and one copy otherwise
+    void ValidateCopyCount(const TextureSpec& textureSpec, const BufferSpec& bufferSpec, const CopySplit& copySplit) {
```
Kangz: This is very tied to the implementation detail of the split, and the number of copies is kind of an implementation detail (not important for correctness, only for perf). So I think we could remove this check.

src/tests/unittests/d3d12/CopySplit.cpp
```diff
@@ -0,0 +1,375 @@
...
+            return copySplit;
+        }
+};
+
+TEST_F(CopySplitTest, General) {
+    for (TextureSpec textureSpec : kBaseTextureSpecs) {
+        for (BufferSpec bufferSpec : BaseBufferSpecs(textureSpec)) {
+
+            CopySplit copySplit = DoTest(textureSpec, bufferSpec);
+            if (HasFatalFailure()) {
```
Kangz: What are these blocks for? It could be factored into the DoTest function.

src/backend/d3d12/TextureCopySplitter.h
```diff
@@ -0,0 +1,54 @@
...
+    constexpr unsigned int kMaxTextureCopyRegions = 2;
+
+    struct CopyInfo {
+        struct Region {
+            uint32_t x = 0;
+            uint32_t y = 0;
+            uint32_t z = 0;
+            uint32_t width;
+            uint32_t height;
+            uint32_t depth;
```
kainino0x: not that important, but may as well zero these 3 fields as well?

src/backend/d3d12/TextureCopySplitter.h
```diff
@@ -0,0 +1,54 @@
...
+        };
+
+        Region texture;
+        Region buffer;
+    };
+
+    struct CopySplit {
+        uint32_t offset;
+        uint32_t count;
```
kainino0x: I would zero the offset and count here too.

src/backend/d3d12/TextureCopySplitter.h
```diff
@@ -0,0 +1,54 @@
...
+#ifndef BACKEND_D3D12_TEXTURECOPYSPLITTER_H_
+#define BACKEND_D3D12_TEXTURECOPYSPLITTER_H_
+
+#include "nxt/nxtcpp.h"
+
+#include <array>
+
+namespace backend {
+namespace d3d12 {
+namespace TextureCopy {
```
austinEng: Done

src/backend/d3d12/TextureCopySplitter.h
```diff
@@ -0,0 +1,54 @@
...
+        struct Region {
+            uint32_t x = 0;
+            uint32_t y = 0;
+            uint32_t z = 0;
+            uint32_t width;
+            uint32_t height;
+            uint32_t depth;
+        };
+
+        Region texture;
```
austinEng: Done

src/backend/d3d12/TextureCopySplitter.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+#include "backend/d3d12/d3d12_platform.h"
+#include "common/Assert.h"
+
+namespace backend {
+namespace d3d12 {
+
+    namespace {
+        void ComputeTexelOffsets(uint32_t offset, uint32_t rowPitch, uint32_t slicePitch, uint32_t texelSize, uint32_t* texelOffsetX, uint32_t* texelOffsetY, uint32_t* texelOffsetZ) {
+            uint32_t byteOffsetZ = (offset / slicePitch) * slicePitch;
```
austinEng: Done

src/backend/d3d12/TextureCopySplitter.cpp
```diff
@@ -0,0 +1,192 @@
...
+            // TODO(enga@google.com): Handle 3D / 2D arrays
+            ASSERT(false);
+            return copy;
+        }
+
+        ASSERT(rowPitch % texelSize == 0);
+
+        uint32_t alignedOffset = offset & ~(D3D12_TEXTURE_DATA_PLACEMENT_ALIGNMENT - 1);
+
+        copy.offset = alignedOffset;
```
austinEng: Done

src/backend/d3d12/TextureCopySplitter.cpp
```diff
@@ -0,0 +1,192 @@
...
+                ASSERT(width > copy.copies[0].texture.width);
+                copy.copies[1].texture.width = width - copy.copies[0].texture.width;
+                copy.copies[1].texture.height = height;
+                copy.copies[1].texture.depth = depth;
+
+                copy.copies[1].buffer.x = 0;
+                copy.copies[1].buffer.y = texelOffsetY + 1;
+                copy.copies[1].buffer.z = texelOffsetZ;
+                ASSERT(texelOffsetX + width > rowPitch / texelSize);
+                copy.copies[1].buffer.width = width + texelOffsetX - rowPitch / texelSize;
```
austinEng: Done

src/tests/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ set(UNITTESTS_DIR ${TESTS_DIR}/unittests)
 set(VALIDATION_TESTS_DIR ${UNITTESTS_DIR}/validation)
 set(END2END_TESTS_DIR ${TESTS_DIR}/end2end)
 
-add_executable(nxt_unittests
+list(APPEND TEST_SOURCES
```
austinEng: Done

src/tests/CMakeLists.txt
```diff
@@ -53,6 +53,14 @@ add_executable(nxt_unittests
     ${VALIDATION_TESTS_DIR}/ValidationTest.h
     ${TESTS_DIR}/UnittestsMain.cpp
 )
+
+if (NXT_ENABLE_D3D12)
+    list(APPEND TEST_SOURCES
+        ${UNITTESTS_DIR}/d3d12/CopySplit.cpp
```
austinEng: Done

src/tests/unittests/d3d12/CopySplit.cpp
```diff
@@ -0,0 +1,375 @@
...
+        }
+    }
+
+    // Check that the offset is aligned
+    void ValidateOffset(const CopySplit& copySplit) {
+        ASSERT_TRUE(Align(copySplit.offset, D3D12_TEXTURE_DATA_PLACEMENT_ALIGNMENT) == copySplit.offset);
+    }
+
+    bool RangesOverlap(uint32_t minA, uint32_t maxA, uint32_t minB, uint32_t maxB) {
+        return (minA < minB && minB >= maxA) || (minB < minA && minA >= maxB);
```
austinEng: Done

src/tests/unittests/d3d12/CopySplit.cpp
austinEng: Done

src/tests/unittests/d3d12/CopySplit.cpp
austinEng: Done

src/tests/unittests/d3d12/CopySplit.cpp
austinEng: > check that for each copy the first pixel is correctly copied

Might be misunderstanding, but not sure how I would actually test this without re-implementing the offset computation. Checking the first pixel of each copy seems like it would require projecting the buffer offset to a texel offset the same way it's done in TextureCopySplitter

src/tests/unittests/d3d12/CopySplit.cpp
austinEng: I want to print which split failed, and then if so exit the test. If I don't exit immediately, it's going to print for every failure after the first failure. If I factor it out, then the FAIL() macro won't exit the test because the result of the FAIL will be wrapped in another function

src/backend/d3d12/TextureCopySplitter.h
austinEng: Done

src/backend/d3d12/TextureCopySplitter.h
austinEng: Done

src/tests/unittests/d3d12/CopySplit.cpp
Kangz: You could do it by doing something like this:

```
// For pixel (0, 0, 0) of the copy
uint32_t absoluteTexelOffset = copies.offset + rowPitch * (bufferOffset.y + bufferOffset.z * bufferSize.z) + bufferOffset.x;
uint32_t relativeTexelOffset = copies.offset - spec.bufferOffset;
uint32_t x = relativeTexelOffset % rowPitchInTexels;
relativeTexelOffset -= x;
// This only works for 2D, will also need handling of Z.
uint32_t y = relativeTexelOffset / rowPitchInTexels;

ASSERT_EQ(spec.x + textureOffset.x, x);
// Same for Y, Z.
```

src/tests/unittests/d3d12/CopySplit.cpp
Kangz: TIL, I thought FAIL() used exceptions. Sounds good.

src/tests/unittests/d3d12/CopySplit.cpp
austinEng: Done

Kangz: LGTM

PR #91, opened 2017-07-21
clang/gcc: enable a bunch more warnings

kainino0x:
I don't feel that strongly about keeping all of these. However, it works right now. Please feel free to remove some later on if they turn out to be too annoying.

(we can also remove some from this PR based on review)


src/tests/unittests/CommandAllocatorTests.cpp
```diff
@@ -260,14 +260,15 @@ TEST(CommandAllocator, ManySmallCommands) {
     iterator.DataWasDestroyed();
 }
 
-//        ________
-//       /        \
-//       | POUIC! |
-//       \_ ______/
-//         v
```
Kangz: /pokerface

src/tests/end2end/BasicTests.cpp
```diff
@@ -34,4 +34,66 @@ TEST_P(BasicTests, BufferSetSubData) {
     EXPECT_BUFFER_U32_EQ(value, buffer, 0);
 }
 
+TEST_P(BasicTests, ReadPixelsTest) {
```
Kangz: This change somehow snuck in but has been removed since then.

kainino0x: argh, rebase error. I thought I fixed it.

PR #94, opened 2017-07-25
Swap chains, part 2

kainino0x:
* [x] D3D12 implementation
* [x] Fix D3D12 error in end2end tests
* [ ] ~~Move swap chain implementations out of bindings into SampleUtils if possible?~~

### DIFF
--- a/src/tests/NXTTest.cpp
+++ b/src/tests/NXTTest.cpp
@@ -137,6 +137,10 @@ void NXTTest::SetUp() {
     device = nxt::Device::Acquire(backendDevice);
     queue = device.CreateQueueBuilder().GetResult();
 
+    swapchain = device.CreateSwapChainBuilder()
+        .SetImplementation(binding->GetSwapChainImplementation())
+        .GetResult();
+
     device.SetErrorCallback(DeviceErrorCauseTestFailure, 0);
 }
 
@@ -215,6 +219,13 @@ std::ostringstream& NXTTest::AddTextureExpectation(const char* file, int line, c
 void NXTTest::WaitABit() {
     device.Tick();
     utils::USleep(100);
+}
+
+void NXTTest::SwapBuffersForCapture() {
+    // Insert a frame boundary for API capture tools.
+    nxt::Texture backBuffer = swapchain.GetNextTexture();
+    backBuffer.TransitionUsage(nxt::TextureUsageBit::Present);
+    swapchain.Present(backBuffer);
 }
 
 NXTTest::ReadbackReservation NXTTest::ReserveReadback(uint32_t readbackSize) {
--- a/src/tests/NXTTest.cpp
+++ b/src/tests/NXTTest.cpp
@@ -217,10 +217,6 @@ void NXTTest::WaitABit() {
     utils::USleep(100);
 }
 
-void NXTTest::SwapBuffers() {
-    binding->SwapBuffers();
-}
-
 NXTTest::ReadbackReservation NXTTest::ReserveReadback(uint32_t readbackSize) {
     // For now create a new MapRead buffer for each readback
     // TODO(cwallez@chromium.org): eventually make bigger buffers and allocate linearly?
--- a/src/utils/D3D12Binding.cpp
+++ b/src/utils/D3D12Binding.cpp
@@ -16,6 +16,7 @@
 
 #include "common/Assert.h"
 #include "nxt/nxt_wsi.h"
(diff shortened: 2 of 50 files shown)


Kangz:
The swapchai implementation for GLFW should stay in utils/ imho because it is also useful for things like tests.

kainino0x:
OK.

kainino0x:
Sorry for enormous patch. I think this is finally done (except maybe end2end test failures, not sure yet). @Kangz, PTAL; @austinEng, PTAL for the D3D12 changes.

src/tests/NXTTest.cpp
```diff
@@ -217,10 +217,6 @@ void NXTTest::WaitABit() {
     utils::USleep(100);
 }
 
-void NXTTest::SwapBuffers() {
```
Kangz: Can we keep an equivalent of this like

```
void SwapBuffersForCapture() {
    nxt::Texture backBuffer = swapchain.GetNextTexture();
    backBuffer.TransitionUsage(nxt::TextureUsageBit::Present);
    swapchain.Present(backBuffer);
}
```

src/utils/MetalBinding.mm
```diff
@@ -30,49 +33,154 @@
...
+            }
+            void GetProcAndDevice(nxtProcTable* procs, nxtDevice* device) override {
+                metalDevice = MTLCreateSystemDefaultDevice();
+
+                backend::metal::Init(metalDevice, procs, device);
+                backendDevice = *device;
+            }
+
+            uint64_t GetSwapChainImplementation() override {
+				if (swapchainImpl.userData == nullptr) {
```
Kangz: nit: indentation

src/utils/D3D12Binding.cpp
```diff
@@ -188,32 +218,70 @@ namespace utils {
...
                 lastSerialRenderTargetWasUsed[renderTargetIndex] = backend::d3d12::GetSerial(backendDevice);
 
-                // Tell the backend to render to the current render target
-                backend::d3d12::SetNextTexture(backendDevice, renderTargetResources[renderTargetIndex]);
+                return NXT_SWAP_CHAIN_NO_ERROR;
+            }
+
+            // C stubs for C++ methods
+
+            static void Init(void* userData, void* wsiContext) {
```
Kangz: nit: we could have a templated helper that handles making these function and filling the nxtSwapChainImplementation so that we avoid doing it in every Binding file.

kainino0x:
Oh right, you mentioned that issue and I completely forgot. Thanks for the details, I think I understand the issue now.

src/utils/MetalBinding.mm
```diff
@@ -30,49 +33,154 @@
...
+            nxtSwapChainError GetNextTexture(nxtSwapChainNextTexture* nextTexture) {
+                [currentDrawable release];
+                currentDrawable = [layer nextDrawable];
+                [currentDrawable retain];
+
+                [currentTexture release];
+                currentTexture = currentDrawable.texture;
+                [currentTexture retain];
+
+                // Clear initial contents of the texture
```
Kangz: Instead of this, can we make BeginSubpass in CommandBufferMTL default to MTLLoadActionClear for color attachments? It might break the RenderToTexture sample but that's fine because we'll add load ops next.

src/utils/MetalBinding.mm
```diff
@@ -30,49 +33,154 @@
...
+				if (swapchainImpl.userData == nullptr) {
+					swapchainImpl = SwapChainImplMTL::Create(glfwGetCocoaWindow(window));
+				}
+				return reinterpret_cast<uint64_t>(&swapchainImpl);
+            }
+
+        private:
+            id<MTLDevice> metalDevice = nil;
             nxtDevice backendDevice = nullptr;
+			nxtSwapChainImplementation swapchainImpl = {};
```
Kangz: nit: indentation

src/utils/OpenGLBinding.cpp
```diff
@@ -14,20 +14,130 @@
...
+            SwapChainImplGL(GLFWwindow* window)
+                : window(window) {
+            }
+
+            ~SwapChainImplGL() {
+                glDeleteTextures(1, &backTexture);
+                glDeleteFramebuffers(1, &backFBO);
+            }
+
+            void HACKCLEAR() {
```
Kangz: Same thing as for the Metal swapchain, let's remove hackClear and do it in BeginSubpass instead.

PR #95, opened 2017-07-27
D3D12: Improve live object reporting and actually clean up resources

austinEng:
This adds better live object reporting in debug mode and fixes a bunch of memory leaks which were happening because I never properly deleted things. Now there are no longer 927 live objects after running the end2end tests :P

I think there are still a couple problems though. It's reporting a ID3D12RootSignature as live when some of the tests exit.


austinEng:
It fixes nearly all of them. There still seems to be live `ID3D12RootSignature`s

PR #98, opened 2017-07-28
.gitignore + make in-tree builds work

kainino0x:
This somewhat reduces noise for in-tree builds, too, in case any contributor wants to or accidentally does that.


.gitignore
```diff
@@ -0,0 +1,119 @@
+generator/
```
Kangz: This should be more specific as this will ignore files added to templates etc.

.gitignore
```diff
@@ -1,4 +1,10 @@
-generator/
+generator/mock
```
Kangz: Can we change the directory where things are generated instead to be like "generator/generated" ?

kainino0x:
I was trying to test the .gitignore for in-tree builds and realized the in-tree build actually doesn't work on mac/linux because of a filename conflict. Pushed a fix.

However this actually causes existing partial builds to break if you don't remove build/examples/glTFViewer first (because the new output file is build/examples/glTFViewer/glTFViewer). What's your preference on allowing in-tree builds? I could just omit that change.

kainino0x:
np, can you mark as approved?

PR #99, opened 2017-07-28
Initial changes for depth stencil in D3D12 backend

BenConMS:
Here are the discussed changes. I will still be looking at the code more to see what I can find to figure out the failing cases, but any discussion in the meantime where code inspection finds the issues is most welcome.


src/backend/d3d12/DepthStencilStateD3D12.cpp
```diff
@@ -0,0 +1,101 @@
...
+// limitations under the License.
+
+#include "backend/d3d12/DepthStencilStateD3D12.h"
+
+#include "common/BitSetIterator.h"
+
+namespace backend {
+namespace d3d12 {
+
+	static D3D12_STENCIL_OP StencilOp(nxt::StencilOperation op) {
```
Kangz: nit: for some reason we use anonymous namespaces instead of marking functions static.

src/backend/d3d12/DepthStencilStateD3D12.cpp
```diff
@@ -0,0 +1,101 @@
...
+		switch (op) {
```
Kangz: Here and in the rest of the PR, we use 4 spaces for indents instead of spaces.

src/backend/d3d12/DepthStencilStateD3D12.cpp
```diff
@@ -0,0 +1,101 @@
...
+		desc.StencilFunc = ComparisonFunc(faceInfo.compareFunction);
+
+		return desc;
+	}
+
+	DepthStencilState::DepthStencilState(Device* device, DepthStencilStateBuilder* builder)
+		: DepthStencilStateBase(builder), device(device) {
+
+		// If you have anything other than Never, then enable depth testing
+		depthStencilDescriptor.DepthEnable = TRUE; // (GetDepth().compareFunction == nxt::CompareFunction::Never) ? FALSE : TRUE;
```
Kangz: Do test work with the commented out code?

src/backend/d3d12/DepthStencilStateD3D12.cpp
```diff
@@ -0,0 +1,101 @@
...
+		depthStencilDescriptor.DepthWriteMask = GetDepth().depthWriteEnabled ? D3D12_DEPTH_WRITE_MASK_ALL : D3D12_DEPTH_WRITE_MASK_ZERO;
+		depthStencilDescriptor.DepthFunc = ComparisonFunc(GetDepth().compareFunction);
+
+		depthStencilDescriptor.StencilEnable = StencilTestEnabled() ? TRUE : FALSE;
+		depthStencilDescriptor.StencilReadMask = (UINT8)GetStencil().readMask;
```
Kangz: nit: C++ static_cast and reinterpret_cast instead of C-style casts.

src/backend/d3d12/GeneratedCodeIncludes.h
```diff
@@ -20,6 +20,7 @@
 #include "backend/d3d12/ComputePipelineD3D12.h"
 #include "backend/d3d12/FramebufferD3D12.h"
 #include "backend/d3d12/InputStateD3D12.h"
+#include "backend/d3d12/DepthStencilStateD3D12.h"
```
Kangz: nit: order

src/backend/d3d12/RenderPipelineD3D12.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include "backend/d3d12/D3D12Backend.h"
 #include "backend/d3d12/InputStateD3D12.h"
+#include "backend/d3d12/DepthStencilStateD3D12.h"
```
Kangz: nit:order

src/tests/end2end/DepthStencilStateTests.cpp
```diff
@@ -522,4 +522,4 @@ TEST_P(DepthStencilStateTest, StencilDepthPass) {
 2);                                                         // Replace the stencil on stencil pass, depth pass, so it should be 2
 }
 
-NXT_INSTANTIATE_TEST(DepthStencilStateTest, MetalBackend, OpenGLBackend)
+NXT_INSTANTIATE_TEST(DepthStencilStateTest, MetalBackend, OpenGLBackend, D3D12Backend)
```
Kangz: nit: order

src/backend/d3d12/RenderPipelineD3D12.cpp
```diff
@@ -120,12 +121,14 @@ namespace d3d12 {
...
-        descriptor.DepthStencilState.DepthEnable = false;
-        descriptor.DepthStencilState.StencilEnable = false;
+		DepthStencilState* depthStencilState = ToBackend(GetDepthStencilState());
+		descriptor.DepthStencilState = depthStencilState->GetD3D12DepthStencilDescriptor();
+
         descriptor.SampleMask = UINT_MAX;
         descriptor.PrimitiveTopologyType = D3D12_PRIMITIVE_TOPOLOGY_TYPE_TRIANGLE;
         descriptor.NumRenderTargets = 1;
         descriptor.RTVFormats[0] = DXGI_FORMAT_R8G8B8A8_UNORM;
+        descriptor.DSVFormat = DXGI_FORMAT_D32_FLOAT_S8X24_UINT;
```
Kangz: This change LGTM, the following can be addressed in a follow-up patch.

NumRenderTargets, RTVFormats and DSVFormats should depend on the information that has been set for the (renderpass, subpass) for this pipeline. Right now the Debug layer will complain in the many cases where we don't have a DSV set in OMSetRenderTargets. Also @kainino0x we should make sure we include some multiple attachment tests when testing render passes.

src/backend/d3d12/DepthStencilStateD3D12.cpp
BenConMS: Not sure what you mean here. I copied this style from InputStateD3D12.cpp.

By anonymous namespace do you mean not scoping inside a namespace at all?

src/backend/d3d12/DepthStencilStateD3D12.cpp
BenConMS: You mean instead of tabs? We use spaces here too, but a fresh VS install always means removing tabs! Sorry about that.

src/backend/d3d12/DepthStencilStateD3D12.cpp
Kangz: Doing something like

```
namespace {
    D3D12_STENCIL_OP StencilOp(nxt::StencilOperation op) {
         // stuff 
    }

    // Other file-scoped functions.
}
```

src/backend/d3d12/DepthStencilStateD3D12.cpp
Kangz: Yeah I meant space instead of tabs. No worries :)

src/backend/d3d12/DepthStencilStateD3D12.cpp
```diff
@@ -0,0 +1,101 @@
...
+        desc.StencilFunc = ComparisonFunc(faceInfo.compareFunction);
+
+        return desc;
+    }
+
+    DepthStencilState::DepthStencilState(Device* device, DepthStencilStateBuilder* builder)
+        : DepthStencilStateBase(builder), device(device) {
+
+        // If you have anything other than Never, then enable depth testing
+        depthStencilDescriptor.DepthEnable = TRUE;
```
Kangz: Would ```(GetDepth().compareFunction == nxt::CompareFunction::Never) ? FALSE : TRUE;``` speed things up in some cases?

Kangz:
Thanks for the update, it looks ready (one question). We just need to figure out this CLA thibg.

src/backend/d3d12/DepthStencilStateD3D12.cpp
BenConMS: It is difficult to answer if it would speed things up across all of the different hardware vendors and generations. There might be some theoretical boost but I think that binding / unbinding a depth buffer is probably already going to be a big hint that they follow.

After some internal discussion here I think that the NXT API needs to be explicit here. There are a few problems with implying the disable of depth checking in this way:

a) It is not super clear from the outside that you disable the hardware (including stencil and depth write, etc) just from setting one part of the depth stencil description.

b) If we try to do that here we will have two backends that are trying to extract real hardware state from a combination of state inputs.

Based on some of the previous remarks about this topic I think at a minimum we should revisit the API before we look for optimization opportunities here. Should we make an issue for this?

src/backend/d3d12/RenderPipelineD3D12.cpp
```diff
@@ -120,12 +121,14 @@ namespace d3d12 {
...
-        descriptor.DepthStencilState.DepthEnable = false;
-        descriptor.DepthStencilState.StencilEnable = false;
+        DepthStencilState* depthStencilState = ToBackend(GetDepthStencilState());
+        descriptor.DepthStencilState = depthStencilState->GetD3D12DepthStencilDescriptor();
+
         descriptor.SampleMask = UINT_MAX;
         descriptor.PrimitiveTopologyType = D3D12_PRIMITIVE_TOPOLOGY_TYPE_TRIANGLE;
         descriptor.NumRenderTargets = 1;
         descriptor.RTVFormats[0] = DXGI_FORMAT_R8G8B8A8_UNORM;
+        descriptor.DSVFormat = DXGI_FORMAT_D32_FLOAT_S8X24_UINT;
```
RafaelCintron: In the future, we should consider putting DXGI_FORMAT_R8G8B8A8_UNORM and DXGI_FORMAT_D32_FLOAT_S8X24_UINT into a common place so that texture creation and descriptor setup do not accidentally get misconfigured from one another.

Kangz:
Message to make CLA bot recheck the CLA (hopefully).

src/backend/d3d12/RenderPipelineD3D12.cpp
Kangz: Yes that's the interesting issue outlined above, we should gather the attachment format from the renderpass and subpass. At nxt::Framebuffer creation we will (or do already?) check that the framebuffer texture formats match the ones of the renderpass, so everything will line up correctly.

austinEng: Yes, we do check they have the same format.
`d3d12::Texture` is also checking the format when it creates the render target view descriptors and `d3d12::Framebuffer` checks if it's a depth texture so the render target / depth stencil views are created in the right descriptor heap

src/backend/d3d12/DepthStencilStateD3D12.cpp
Kangz: Thanks for the data. I don't share your point of view on this, as I believe that enabling or disabling depth write is redundant information, and that the application should just have to declare its intent. That said, this is an interesting question that we should discuss later with the wider GPUWeb group. Let's create an issue to remember this, but no do any changes to NXT about it as the API is functional as is. WDYT?

Also the comment above should be removed as it refers to deleted logic.

src/backend/d3d12/RenderPipelineD3D12.cpp
Kangz: Also original comment about this that is hidden inside Github's comments:

> This change LGTM, the following can be addressed in a follow-up patch.
>
>NumRenderTargets, RTVFormats and DSVFormats should depend on the information that has been set for the (renderpass, subpass) for this pipeline. Right now the Debug layer will complain in the many cases where we don't have a DSV set in OMSetRenderTargets. Also @kainino0x we should make sure we include some multiple attachment tests when testing render passes.

src/backend/d3d12/DepthStencilStateD3D12.cpp
BenConMS: I agree on filing an issue and not changing NXT. 

I removed the comment since it no longer made sense to be there.

PR #103, opened 2017-08-02
Blend state

austinEng:
Not working yet.
 - As usual, this is completely untested on Metal. But for some reason it's saying some enums which I got from the docs aren't defined?
- This PR also sets the render target formats based on the subpass
- Updates `shaderc` so we can have 4 draw buffers
- Needs some comments

### DIFF
--- a/next.json
+++ b/next.json
@@ -105,6 +105,78 @@
             {"value": 3, "name": "storage buffer"}
         ]
     },
+    "blend factor": {
+        "category": "enum",
+        "values": [
+            {"value": 0, "name": "zero"},
+            {"value": 1, "name": "one"},
+            {"value": 2, "name": "src color"},
+            {"value": 3, "name": "one minus src color"},
+            {"value": 4, "name": "src alpha"},
+            {"value": 5, "name": "one minus src alpha"},
+            {"value": 6, "name": "dst color"},
+            {"value": 7, "name": "one minus dst color"},
+            {"value": 8, "name": "dst alpha"},
+            {"value": 9, "name": "one minus dst alpha"},
+            {"value": 10, "name": "src alpha saturated"},
+            {"value": 11, "name": "blend color"},
+            {"value": 12, "name": "one minus blend color"},
+            {"value": 13, "name": "src1 color"},
+            {"value": 14, "name": "one minus src1 color"},
+            {"value": 15, "name": "src1 alpha"},
+            {"value": 16, "name": "one minus src1 alpha"}
+        ]
+    },
+    "blend operation": {
+        "category": "enum",
+        "values": [
+            {"value": 0, "name": "add"},
+            {"value": 1, "name": "subtract"},
+            {"value": 2, "name": "reverse subtract"},
+            {"value": 3, "name": "min"},
+            {"value": 4, "name": "max"}
+        ]
+    },
+    "blend state": {
+        "category": "object"
+    },
+    "blend state builder": {
+        "category": "object",
+        "methods": [
+            {
+                "name": "get result",
+                "returns": "blend state"
+            },
+            {
+                "name": "set blend enabled",
+                "args": [
+                    {"name": "blend enabled", "type": "bool"}
+                ]
+            },
+            {
+                "name": "set alpha blend",
+                "args": [
+                    {"name": "blend operation", "type": "blend operation"},
+                    {"name": "src factor",  "type": "blend factor"},
+                    {"name": "dst factor", "type": "blend factor"}
+                ]
+            },
+            {
+                "name": "set color blend",
+                "args": [
+                    {"name": "blend operation", "type": "blend operation"},
+                    {"name": "src factor",  "type": "blend factor"},
+                    {"name": "dst factor", "type": "blend factor"}
+                ]
+            },
+            {
+                "name": "set color write mask",
+                "args": [
+                    {"name": "color write mask", "type": "color write mask"}
+                ]
+            }
+        ]
+    },
     "builder error status": {
         "category": "enum",
         "values": [
@@ -236,6 +308,17 @@
     },
     "char": {
         "category": "native"
+    },
+    "color write mask": {
+        "category": "bitmask",
+        "values": [
+            {"value": 0, "name": "none"},
+            {"value": 1, "name": "red"},
+            {"value": 2, "name": "green"},
+            {"value": 4, "name": "blue"},
+            {"value": 8, "name": "alpha"},
+            {"value": 15, "name": "all"}
+        ]
     },
     "command buffer": {
         "category": "object"
@@ -472,6 +555,10 @@
             {
                 "name": "create bind group layout builder",
                 "returns": "bind group layout builder"
+            },
+            {
+                "name": "create blend state builder",
+                "returns": "blend state builder"
             },
             {
                 "name": "create buffer builder",
@@ -786,6 +873,13 @@
             {
                 "name": "get result",
                 "returns": "render pipeline"
+            },
+            {
+                "name": "set color attachment blend state",
+                "args": [
+                    {"name": "attachment slot", "type": "uint32_t"},
+                    {"name": "blend state", "type": "blend state"}
+                ]
             },
             {
                 "name": "set depth stencil state",
--- a/next.json
+++ b/next.json
@@ -120,11 +120,7 @@
             {"value": 9, "name": "one minus dst alpha"},
             {"value": 10, "name": "src alpha saturated"},
             {"value": 11, "name": "blend color"},
-            {"value": 12, "name": "one minus blend color"},
-            {"value": 13, "name": "src1 color"},
-            {"value": 14, "name": "one minus src1 color"},
-            {"value": 15, "name": "src1 alpha"},
-            {"value": 16, "name": "one minus src1 alpha"}
+            {"value": 12, "name": "one minus blend color"}
         ]
     },
     "blend operation": {
--- a/next.json
+++ b/next.json
@@ -442,6 +442,15 @@
                 ]
             },
             {
+                "name": "set blend color",
+                "args": [
+                    {"name": "r", "type": "float"},
+                    {"name": "g", "type": "float"},
+                    {"name": "b", "type": "float"},
+                    {"name": "a", "type": "float"}
+                ]
+            },
+            {
                 "name": "set bind group",
                 "args": [
                     {"name": "group index", "type": "uint32_t"},
@@ -687,6 +696,9 @@
             {"value": 0, "name":"nearest"},
             {"value": 1, "name":"linear"}
         ]
+    },
+    "float": {
+        "category": "native"
     },
     "framebuffer": {
         "category": "object"
--- a/src/backend/CommandBuffer.cpp
+++ b/src/backend/CommandBuffer.cpp
@@ -236,6 +236,12 @@ namespace backend {
                         cmd->~SetStencilReferenceCmd();
                     }
                     break;
+                case Command::SetBlendColor:
+                    {
+                        SetBlendColorCmd* cmd = commands->NextCommand<SetBlendColorCmd>();
+                        cmd->~SetBlendColorCmd();
+                    }
+                    break;
                 case Command::SetBindGroup:
                     {
                         SetBindGroupCmd* cmd = commands->NextCommand<SetBindGroupCmd>();
@@ -343,6 +349,10 @@ namespace backend {
 
             case Command::SetStencilReference:
                 commands->NextCommand<SetStencilReferenceCmd>();
+                break;
+
+            case Command::SetBlendColor:
+                commands->NextCommand<SetBlendColorCmd>();
                 break;
 
             case Command::SetBindGroup:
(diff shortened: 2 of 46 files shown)


src/backend/metal/BlendStateMTL.mm
```diff
@@ -0,0 +1,104 @@
...
+                    return MTLBlendOperationReverseSubtract;
+                case nxt::BlendOperation::Min:
+                    return MTLBlendOperationMin;
+                case nxt::BlendOperation::Max:
+                    return MTLBlendOperationMax;
+            }
+        }
+
+        MTLColorWriteMask MetalColorWriteMask(nxt::ColorWriteMask colorWriteMask) {
+            return (
```
Kangz: nit: Metal uses an inverse order compared to NXT so we can't do the same thing as for D3D12

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -106,15 +106,21 @@ namespace opengl {
 
                         const auto& info = currentRenderPass->GetSubpassInfo(currentSubpass);
 
-                        for (uint32_t index = 0; index < info.colorAttachments.size(); ++index) {
-                            uint32_t attachment = info.colorAttachments[index];
+                        GLenum drawBuffers[kMaxColorAttachments];
```
Kangz: This could be an std::array and should be filled with GL_NONE by default.

src/tests/end2end/BlendStateTests.cpp
```diff
@@ -0,0 +1,1055 @@
...
+
+class BlendStateTest : public NXTTest {
+    protected:
+        void SetUp() override {
+            NXTTest::SetUp();
+
+            vsModule = utils::CreateShaderModule(device, nxt::ShaderStage::Vertex, R"(
+                #version 450
+                void main() {
+                    const vec2 pos[3] = vec2[3](vec2(0.0f, 1.0f), vec2(-1.0f, -1.0f), vec2(1.0f, -1.0f));
```
Kangz: nit: you could make the triangle cover the whole square even.

next.json
```diff
@@ -105,6 +105,78 @@
...
+            {"value": 4, "name": "src alpha"},
+            {"value": 5, "name": "one minus src alpha"},
+            {"value": 6, "name": "dst color"},
+            {"value": 7, "name": "one minus dst color"},
+            {"value": 8, "name": "dst alpha"},
+            {"value": 9, "name": "one minus dst alpha"},
+            {"value": 10, "name": "src alpha saturated"},
+            {"value": 11, "name": "blend color"},
+            {"value": 12, "name": "one minus blend color"},
+            {"value": 13, "name": "src1 color"},
```
Kangz: Or keep it because you implemented tests and all for it, but be ready to remove it / make it an extension later.

src/tests/end2end/BlendStateTests.cpp
```diff
@@ -0,0 +1,1055 @@
...
+    }
+
+    // Blend two RGBA8 color values parameterized by the provided factors in the range [0.f, 1.f]
+    RGBA8 mix(const RGBA8& col1, const RGBA8& col2, std::array<float, 4> fac) {
+        float r = static_cast<float>(col1.r) * (1.f - fac[0]) + static_cast<float>(col2.r) * fac[0];
+        float g = static_cast<float>(col1.g) * (1.f - fac[1]) + static_cast<float>(col2.g) * fac[1];
+        float b = static_cast<float>(col1.b) * (1.f - fac[2]) + static_cast<float>(col2.b) * fac[2];
+        float a = static_cast<float>(col1.a) * (1.f - fac[3]) + static_cast<float>(col2.a) * fac[3];
+
+        // Get the integer locations of the resulting 4D RGBA8 hypercube after the blend
```
Kangz: Shouldn't round() be enough? I don't think the GPU does that euclidean distance computation.

src/tests/end2end/BlendStateTests.cpp
```diff
@@ -0,0 +1,1055 @@
...
+
+        queue.Submit(1, &commands);
+
+        EXPECT_PIXEL_RGBA8_EQ(expected0, renderTargets[2], kRTSize / 2, kRTSize / 2) << "Attachment slot 0 using render target 2 should have been " << color0 << " + " << base << " = " << expected0;
+        EXPECT_PIXEL_RGBA8_EQ(expected1, renderTargets[4], kRTSize / 2, kRTSize / 2) << "Attachment slot 1 using render target 4 should have been " << color1 << " - " << base << " = " << expected1;
+        EXPECT_PIXEL_RGBA8_EQ(expected2, renderTargets[3], kRTSize / 2, kRTSize / 2) << "Attachment slot 2 using render target 3 should have been " << color2 << " = " << expected2 << "(no blending)";
+        EXPECT_PIXEL_RGBA8_EQ(expected3, renderTargets[0], kRTSize / 2, kRTSize / 2) << "Attachment slot 3 using render target 0 should have been min(" << color3 << ", " << base << ") = " << expected3;
+    }
+}
+
```
Kangz: Could we have a the following tests too:
 - BlendConstants are set to a known value (0, 0, 0, 0?) at the beginning of each subpass with no inheritance between subpasses.
 - Validation tests for all the validation you added.

next.json
austinEng: ah, I will remove it. This must be why it's not being found in the Metal header

Kangz: It is gated on OSX 10.12 so probably yes. Removing sounds good.

src/tests/end2end/BlendStateTests.cpp
austinEng: > BlendConstants are set to a known value (0, 0, 0, 0?) at the beginning of each subpass with no inheritance between subpasses.

 We want to reset it at the beginning of every subpass? Why?

Kangz: NXT should do the same thing on all APIs but on D3D12 we have state inheritance whereas we don't have any in Metal. 0, 0 ,0 ,0 was a guess, but maybe it should be 1, 1, 1, 1 or something else.

austinEng: Done. Default in D3D12 is 1,1,1,1 and in Metal it's 0,0,0,0. But I've made the default 0,0,0,0 so that by default we do less work. I believe the default is also 0 in Vulkan.

src/tests/end2end/BlendStateTests.cpp
austinEng: Done.

src/tests/end2end/BlendStateTests.cpp
austinEng: Done

src/backend/opengl/CommandBufferGL.cpp
austinEng: Done

src/backend/d3d12/BlendStateD3D12.cpp
```diff
@@ -0,0 +1,111 @@
...
+                    return D3D12_BLEND_OP_MIN;
+                case nxt::BlendOperation::Max:
+                    return D3D12_BLEND_OP_MAX;
+                default:
+                    UNREACHABLE();
+            }
+        }
+
+        uint8_t D3D12RenderTargetWriteMask(nxt::ColorWriteMask colorWriteMask) {
+            return static_cast<uint8_t>(
```
austinEng: Done

src/backend/RenderPass.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include "backend/RenderPass.h"
 
+#include "backend/BlendState.h"
```
austinEng: Done

src/backend/CommandBuffer.cpp
```diff
@@ -570,6 +580,16 @@ namespace backend {
                     }
                     break;
 
+                case Command::SetBlendColor:
+                    {
+                        iterator.NextCommand<SetBlendColorCmd>();
+                        if (!state->HaveRenderPass()) {
```
austinEng: Done

next.json
austinEng: Done

src/backend/CommandBuffer.cpp
```diff
@@ -563,8 +573,18 @@ namespace backend {
                 case Command::SetStencilReference:
                     {
                         iterator.NextCommand<SetStencilReferenceCmd>();
-                        if (!state->HaveRenderPass()) {
-                            HandleError("Can't set stencil reference without an active render pass");
+                        if (!state->HaveRenderPass() && !state->HaveRenderSubpass()) {
```
Kangz: We can check for HaveRenderSubpass only as it implies HaveRenderPass

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -106,15 +106,21 @@ namespace opengl {
 
                         const auto& info = currentRenderPass->GetSubpassInfo(currentSubpass);
 
-                        for (uint32_t index = 0; index < info.colorAttachments.size(); ++index) {
-                            uint32_t attachment = info.colorAttachments[index];
+                        std::array<GLenum, kMaxColorAttachments> drawBuffers = {};
```
Kangz: Can we explicitly fill this with GL_NONE even if it is 0. Or add a comment explaining why the zero init does what we want?

src/tests/unittests/validation/RenderPipelineValidationTests.cpp
```diff
@@ -57,13 +59,26 @@ class RenderPipelineValidationTest : public ValidationTest {
...
+        nxt::BlendState blendState;
         nxt::PipelineLayout pipelineLayout;
 };
 
 // Test cases where creation should succeed
 TEST_F(RenderPipelineValidationTest, CreationSuccess) {
     AddDefaultStates(AssertWillBeSuccess(device.CreateRenderPipelineBuilder()))
         .GetResult();
+
+    AddDefaultStates(AssertWillBeSuccess(device.CreateRenderPipelineBuilder()))
```
Kangz: nit: this one should actually fail because AddDefaultStates sets the layout.

src/tests/unittests/validation/RenderPipelineValidationTests.cpp
```diff
@@ -99,6 +114,72 @@ TEST_F(RenderPipelineValidationTest, CreationMissingProperty) {
...
+        AssertWillBeSuccess(device.CreateRenderPipelineBuilder())
+            .SetSubpass(renderpass, 0)
+            .SetLayout(pipelineLayout)
+            .SetStage(nxt::ShaderStage::Vertex, vsModule, "main")
+            .SetStage(nxt::ShaderStage::Fragment, fsModule, "main")
+            .SetPrimitiveTopology(nxt::PrimitiveTopology::TriangleList)
+            .SetColorAttachmentBlendState(0, blendState)
+            .GetResult();
+
+        // This failes because attachment 1 is not one of the subpass's color attachments
```
Kangz: nit: typo failes.

src/tests/unittests/validation/RenderPipelineValidationTests.cpp
```diff
@@ -99,6 +114,72 @@ TEST_F(RenderPipelineValidationTest, CreationMissingProperty) {
...
+            .AttachmentSetFormat(0, nxt::TextureFormat::R8G8B8A8Unorm)
+            .AttachmentSetFormat(1, nxt::TextureFormat::R8G8B8A8Unorm)
+            .SetSubpassCount(1)
+            .SubpassSetColorAttachment(0, 0, 0)
+            .GetResult();
+
+        auto framebuffer = device.CreateFramebufferBuilder()
+            .SetRenderPass(renderpass)
+            .SetDimensions(640, 480)
+            .SetAttachment(0, colorView)
```
Kangz: nit: this should fail creation because we are missing one attachment. You don't need to add the validation in the patch, but it should SetAttachment(1) too (with a different texture too).

src/tests/end2end/BlendStateTests.cpp
Kangz: Sounds good thanks!

src/backend/CommandBuffer.cpp
austinEng: Done

src/backend/opengl/CommandBufferGL.cpp
austinEng: Done

src/tests/unittests/validation/RenderPipelineValidationTests.cpp
austinEng: Done

src/tests/unittests/validation/RenderPipelineValidationTests.cpp
austinEng: Done

src/tests/unittests/validation/RenderPipelineValidationTests.cpp
austinEng: Done

PR #104, opened 2017-08-08
Use our fork of SPIRV-Cross with HLSL compute patches

Kangz:


Kangz:
Thanks for the reviews!

PR #105, opened 2017-08-08
Load operations

kainino0x:


kainino0x:
There are issues with some models (e.g. Box) in glTFViewer, but I think these were introduced earlier. The behavior is consistent across backends.

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -256,6 +256,36 @@ namespace d3d12 {
                         D3D12_RECT scissorRect = { 0, 0, static_cast<long>(width), static_cast<long>(height) };
                         commandList->RSSetViewports(1, &viewport);
                         commandList->RSSetScissorRects(1, &scissorRect);
+
+                        for (uint32_t slot = 0; slot < currentRenderPass->GetAttachmentCount(); ++slot) {
```
austinEng: Should this be in BeginRenderSubpass?

src/backend/d3d12/FramebufferD3D12.h
```diff
@@ -34,10 +34,13 @@ namespace d3d12 {
                 unsigned int numRTVs = 0;
                 D3D12_CPU_DESCRIPTOR_HANDLE RTVs[kMaxColorAttachments] = {};
                 D3D12_CPU_DESCRIPTOR_HANDLE dsv = {};
+                uint32_t rtvSlots[kMaxColorAttachments] = {};
```
austinEng: nit: Seems like we prefer to use `std::array`. I also probably should have done that on line 35.

src/backend/d3d12/CommandBufferD3D12.cpp
kainino0x: The semantic here is that all of the clears in a render pass are done "before first use" in that render pass. So for opengl/metal, I clear each attachment on the subpass that first uses it, but on d3d12 I just clear everything once at the beginning. However it could be easily changed to the way I did it in opengl/metal. wdyt?

austinEng: Okay right, and we can't do it this way in Metal because the render encoders are different. I think that this is fine then. It's probably more efficient to put all the clears together anyway?

src/backend/d3d12/FramebufferD3D12.h
kainino0x: Started this on my mac but pushed a broken commit so I can finish it on Windows.

src/backend/d3d12/CommandBufferD3D12.cpp
kainino0x: I don't know about the efficiency. If an application uses fewer than all of the attachments in a render pass, there would be unnecessary clears. But I am not particularly worried about this case at least for now.

src/backend/d3d12/FramebufferD3D12.h
kainino0x: Done

src/backend/RenderPass.cpp
```diff
@@ -51,6 +52,28 @@ namespace backend {
         return other == this;
     }
 
+    void RenderPassBase::ComputeAttachmentsFirstSubpasses(std::vector<uint32_t>* attachmentsFirstSubpasses) const {
```
Kangz: nit: this could be precomputed and stored in the attachment info.

src/backend/d3d12/CommandBufferD3D12.cpp
Kangz: +1 on doing it in BeginSubpass because it should be valid to use the attachments as textures in previous subpasses.

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -256,6 +256,36 @@ namespace d3d12 {
...
+                            const auto& info = currentRenderPass->GetAttachmentInfo(slot);
+                            bool hasDepth = TextureFormatHasDepth(info.format);
```
Kangz: nit: could has a TextureFormatHasDepthOrStencil.

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -256,6 +256,36 @@ namespace d3d12 {
...
+                            bool hasStencil = TextureFormatHasStencil(info.format);
+                            if (hasDepth || hasStencil) {
+                                // This is a depth/stencil buffer
+                                D3D12_CLEAR_FLAGS clearFlags = {};
```
Kangz: nity: = 0 would be more clear

src/backend/d3d12/FramebufferD3D12.cpp
```diff
@@ -68,17 +68,25 @@ namespace d3d12 {
...
-            args.dsv = dsvHeap.GetCPUHandle(heapIndex);
+            uint32_t slot = subpassInfo.depthStencilAttachment;
+            args.dsv = GetAttachmentDescriptor(true, slot);
+            args.dsvSlot = slot;
         }
 
         return args;
     }
 
+    D3D12_CPU_DESCRIPTOR_HANDLE Framebuffer::GetAttachmentDescriptor(bool isDSV, uint32_t attachmentSlot) {
```
Kangz: nit: could have two methods instead of the isDSV boolean.

src/backend/d3d12/FramebufferD3D12.cpp
```diff
@@ -68,17 +68,25 @@ namespace d3d12 {
         const auto& subpassInfo = GetRenderPass()->GetSubpassInfo(subpassIndex);
         OMSetRenderTargetArgs args = {};
 
-        for (uint32_t index : IterateBitSet(subpassInfo.colorAttachmentsSet)) {
-            uint32_t heapIndex = attachmentHeapIndices[subpassInfo.colorAttachments[index]];
-            args.RTVs[args.numRTVs++] = rtvHeap.GetCPUHandle(heapIndex);
+        for (uint32_t location : IterateBitSet(subpassInfo.colorAttachmentsSet)) {
+            uint32_t slot = subpassInfo.colorAttachments[location];
+            args.RTVs[args.numRTVs] = GetAttachmentDescriptor(false, slot);
+            args.rtvSlots[args.numRTVs] = slot;
```
Kangz: rtvSlots and dsvSlots are set but never read. They could be removed.

src/backend/metal/CommandBufferMTL.mm
```diff
@@ -87,32 +87,56 @@ void BeginSubpass(id<MTLCommandBuffer> commandBuffer, uint32_t subpass) {
...
+                    auto clearValue = currentFramebuffer->GetClearColor(location);
+                    if (currentRenderPass->GetAttachmentFirstSubpass(attachment) == subpass) {
+                        const auto& attachmentInfo = currentRenderPass->GetAttachmentInfo(attachment);
+                        if (attachmentInfo.colorLoadOp == nxt::LoadOp::Clear) {
+                            load = MTLLoadActionClear;
+                        }
+                    }
+
+                    descriptor.colorAttachments[location].texture = texture;
+                    descriptor.colorAttachments[location].loadAction = MTLLoadActionClear;
```
Kangz: This should be set to ```load```

src/backend/metal/CommandBufferMTL.mm
```diff
@@ -87,32 +87,56 @@ void BeginSubpass(id<MTLCommandBuffer> commandBuffer, uint32_t subpass) {
...
                     auto textureView = currentFramebuffer->GetTextureView(attachment);
                     auto texture = ToBackend(textureView->GetTexture())->GetMTLTexture();
-                    descriptor.colorAttachments[index].texture = texture;
-                    descriptor.colorAttachments[index].loadAction = MTLLoadActionLoad;
-                    descriptor.colorAttachments[index].clearColor = MTLClearColorMake(0.0, 0.0, 0.0, 0.0);
-                    descriptor.colorAttachments[index].storeAction = MTLStoreActionStore;
+
+                    MTLLoadAction load = MTLLoadActionLoad;
+                    auto clearValue = currentFramebuffer->GetClearColor(location);
+                    if (currentRenderPass->GetAttachmentFirstSubpass(attachment) == subpass) {
```
Kangz: How about the following:


```
descriptor.colorAttachments[location].texture = texture;
descriptor.colorAttachments[location].storeAction = MTLStoreActionStore;

if (firstSubpassUsed && startsCleared) {
    auto clearValue = currentFramebuffer->GetClearColor(location);
    descriptor.colorAttachments[location].loadAction = MTLLoadActionClear;
    descriptor.colorAttachments[location].clearColor = MTLClearColorMake(clearValue.color[0], clearValue.color[1], clearValue.color[2], clearValue.color[3]);
} else {
    descriptor.colorAttachments[location].loadAction = MTLLoadActionLoad
}
```

Kangz: Same below for DS

src/backend/RenderPass.cpp
Kangz: Ah I see it is done per-backend. I think it is a generally useful information and could be stored at the RenderPassBase level.

src/backend/metal/MetalBackend.h
```diff
@@ -168,6 +168,11 @@ namespace metal {
         public:
             RenderPass(RenderPassBuilder* builder);
             ~RenderPass();
+
+            uint32_t GetAttachmentFirstSubpass(uint32_t attachmentSlot);
```
Kangz: With the comments about storing this in RenderPassBase, changes to the file would disappear.

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -96,6 +97,8 @@ namespace opengl {
                 case Command::BeginRenderSubpass:
                     {
                         commands.NextCommand<BeginRenderSubpassCmd>();
+
+                        glBindFramebuffer(GL_READ_FRAMEBUFFER, 0);
```
Kangz: Why do we need to do this?

src/backend/opengl/RenderPassGL.h
```diff
@@ -0,0 +1,34 @@
...
+#define BACKEND_OPENGL_RENDERPASSGL_H_
+
+#include "backend/RenderPass.h"
+
+namespace backend {
+namespace opengl {
+    class RenderPass : public RenderPassBase {
+        public:
+            RenderPass(RenderPassBuilder* builder);
+
```
Kangz: Same here, this specialization of RenderPasses for GL might disappear.

src/tests/unittests/validation/FramebufferValidationTests.cpp
```diff
@@ -77,6 +77,13 @@ TEST_F(FramebufferValidationTest, BasicWithEmptyAttachment) {
         .SetRenderPass(renderpass)
         .SetDimensions(100, 100)
         .GetResult();
+
+    // set clear color value for attachment 0
+    framebuffer.AttachmentSetClearColor(0, 0.0f, 0.0f, 0.0f, 0.0f);
```
Kangz: This test shouldn't be valid anymore and we should force framebuffers to have all attachment defined. Also we should test setting clear colors in a different test.

src/tests/unittests/validation/RenderPassValidationTests.cpp
```diff
@@ -37,6 +37,63 @@ TEST_F(RenderPassValidationTest, OneSubpassOneAttachment) {
...
+        .AttachmentSetFormat(0, nxt::TextureFormat::R8G8B8A8Unorm)
+        // with a load op of the wrong type - this is okay, just ignored
+        .AttachmentSetDepthStencilLoadOps(0, nxt::LoadOp::Clear, nxt::LoadOp::Clear)
+        .SubpassSetColorAttachment(0, 0, 0)
+        .GetResult();
+}
+
+// Test for attachment slot arguments out of bounds
+TEST_F(RenderPassValidationTest, AttachmentOutOfBounds) {
+    AssertWillBeError(device.CreateRenderPassBuilder())
```
Kangz: Can we have a control case that passes, then tests for each OOB possiblity with comments to explain what we are testing? It makes following what the test does easier.

src/tests/unittests/validation/FramebufferValidationTests.cpp
kainino0x: Done but removed from this PR

kainino0x:
Squashed stuff to reduce churn on tests, d3d12's BeginRenderPass, etc.

Tests removed from this PR, coming in a followup PR along with more tests.

src/backend/opengl/CommandBufferGL.cpp
kainino0x: Added a comment but forgot to put it in this PR. Don't want to restart the CI, so I'll commit it in the next PR.

src/backend/metal/CommandBufferMTL.mm
```diff
@@ -87,33 +87,61 @@ void BeginSubpass(id<MTLCommandBuffer> commandBuffer, uint32_t subpass) {
...
 
                     auto textureView = currentFramebuffer->GetTextureView(attachment);
                     auto texture = ToBackend(textureView->GetTexture())->GetMTLTexture();
-                    descriptor.colorAttachments[index].texture = texture;
-                    descriptor.colorAttachments[index].loadAction = MTLLoadActionLoad;
-                    descriptor.colorAttachments[index].clearColor = MTLClearColorMake(0.0, 0.0, 0.0, 0.0);
-                    descriptor.colorAttachments[index].storeAction = MTLStoreActionStore;
+
+                    bool isFirstUse = attachmentInfo.firstSubpass == subpass;
+                    bool isClearedOnFirstUse = attachmentInfo.colorLoadOp == nxt::LoadOp::Clear;
```
Kangz: nit: shouldClearOnFirstUse

src/backend/d3d12/CommandBufferD3D12.cpp
Kangz: Weird, LGTM

kainino0x:
> and removing the spirv-cross submodule change.

argh, I've removed and somehow reintroduced this like 10 times now! Submodules!!!

kainino0x:
@Kangz something seems to be wrong with the linux builder bot, jinja2 is missing. Maybe relevant:
>  This job ran on our Trusty environment, which is gradually becoming our default Linux environment. Read all about this in our blog: Trusty as a default Linux is coming and take note that you can add dist: precise in your .travis.yml file to continue using Precise.

PR #107, opened 2017-08-10
Add tests for clears, render passes, framebuffers

kainino0x:
~~Includes commits from #105. Will rebase once it lands.~~

### DIFF
--- a/src/tests/NXTTest.h
+++ b/src/tests/NXTTest.h
@@ -31,7 +31,7 @@
     AddTextureExpectation(__FILE__, __LINE__, texture, x, y, 1, 1, 0, sizeof(RGBA8), new detail::ExpectEq<RGBA8>(expected))
 
 #define EXPECT_TEXTURE_RGBA8_EQ(expected, texture, x, y, width, height, level) \
-    AddTextureExpectation(__FILE__, __LINE__, texture, x, y, width, height, level, sizeof(RGBA8), new detail::ExpectEq<RGBA8>(expected, width * height))
+    AddTextureExpectation(__FILE__, __LINE__, texture, x, y, width, height, level, sizeof(RGBA8), new detail::ExpectEq<RGBA8>(expected, (width) * (height)))
 
 struct RGBA8 {
     constexpr RGBA8() : RGBA8(0,0,0,0) {}
--- a/src/tests/NXTTest.h
+++ b/src/tests/NXTTest.h
@@ -130,19 +130,6 @@ class NXTTest : public ::testing::TestWithParam<BackendType> {
         utils::BackendBinding* binding = nullptr;
 };
 
-class DrawQuad {
-    public:
-        DrawQuad() {}
-        DrawQuad(nxt::Device* device, const char* vsSource, const char* fsSource);
-        void Draw(const nxt::RenderPass& renderpass, nxt::CommandBufferBuilder* builder);
-
-    private:
-        nxt::Device* device = nullptr;
-        nxt::ShaderModule vsModule = {};
-        nxt::ShaderModule fsModule = {};
-        nxt::PipelineLayout pipelineLayout = {};
-};
-
 // Instantiate the test once for each backend provided after the first argument. Use it like this:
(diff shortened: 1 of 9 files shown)


src/tests/unittests/validation/RenderPassValidationTests.cpp
```diff
@@ -37,6 +37,119 @@ TEST_F(RenderPassValidationTest, OneSubpassOneAttachment) {
         .SetSubpassCount(1)
         .SetAttachmentCount(1)
         .AttachmentSetFormat(0, nxt::TextureFormat::R8G8B8A8Unorm)
+        // without a load op
+        .SubpassSetColorAttachment(0, 0, 0)
+        .GetResult();
+
+    AssertWillBeSuccess(device.CreateRenderPassBuilder())
```
Kangz: nit: can the LoadOp test be in a separate TEST_F?

src/tests/unittests/validation/RenderPassValidationTests.cpp
```diff
@@ -37,6 +37,119 @@ TEST_F(RenderPassValidationTest, OneSubpassOneAttachment) {
...
+    AssertWillBeError(device.CreateRenderPassBuilder())
+        .SetSubpassCount(1)
+        .SetAttachmentCount(1)
+        .AttachmentSetFormat(0, nxt::TextureFormat::D32FloatS8Uint)
+        // Test SubpassSetDepthStencilAttachment subpass out of bounds
+        .SubpassSetDepthStencilAttachment(1, 0)
+        .GetResult();
+}
+
+TEST_F(RenderPassValidationTest, SubpassAttachmentWrongFormat) {
```
Kangz: nit: comment as to what this tests before TEST_F. How about calling the test SubpassAttachmentWrongAspect?

src/tests/end2end/RenderPassLoadOpTests.cpp
```diff
@@ -0,0 +1,179 @@
...
+        .BeginRenderPass(renderpass1, framebuffer1)
+        .BeginRenderSubpass()
+            // Clear should occur implicitly
+            // Store should occur implicitly
+        .EndRenderSubpass()
+        .EndRenderPass()
+        .GetResult();
+
+    // Initialized to 0 before the command buffer is submitted
+    EXPECT_TEXTURE_RGBA8_EQ(expectZero.data(), renderTarget, 0, 0, kRTSize, kRTSize, 0);
```
Kangz: This isn't true yet on any backend, and would be something enabled by browsers when creating the device. It could be replaced with a first clear to black, then a clear to green.

src/tests/end2end/RenderPassLoadOpTests.cpp
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    queue.Submit(1, &commands2);
+    // Left half should still be green
+    EXPECT_TEXTURE_RGBA8_EQ(expectGreen.data(), renderTarget, 0, 0, kRTSize / 2, kRTSize, 0);
+    // Right half should now be blue
+    EXPECT_TEXTURE_RGBA8_EQ(expectBlue.data(), renderTarget, kRTSize / 2, 0, kRTSize / 2, kRTSize, 0);
+}
+
+TEST_P(RenderPassLoadOpTests, LoadFromUninitialized) {
```
Kangz: nit: same comment as above, this shouldn't work for now. I don't think we'll do it in the short term.

src/tests/NXTTest.h
```diff
@@ -130,6 +130,19 @@ class NXTTest : public ::testing::TestWithParam<BackendType> {
         utils::BackendBinding* binding = nullptr;
 };
 
+class DrawQuad {
+    public:
+        DrawQuad() {}
+        DrawQuad(nxt::Device* device, const char* vsSource, const char* fsSource);
```
Kangz: DrawQuad seems like it isn't doing enough for now, we should be able to ask where the quad should be drawn, and only pass the fs for example.

src/tests/end2end/RenderPassLoadOpTests.cpp
kainino0x: Done

PR #111, opened 2017-08-12
Update .travis.yml for trusty

kainino0x:
Not sure if this works yet, trying it out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
  apt:
   sources:
    # PPAs to get newer toolchains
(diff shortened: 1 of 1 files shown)


kainino0x:
Putting this on hold for now:

> NOTE: The allowed list of packages that may be installed varies between Precise and Trusty, with the Precise list being considerably larger at the time of this writing. If the packages you need are not yet available on Trusty, it is recommended that you either target dist: precise or sudo: required depending on your needs.

`sudo: required` is slower so for the moment we'll stick with `dist: precise`.

PR #115, opened 2017-08-14
Command buffer validation fixes

austinEng:
1. `inputsSet` were not being reset at the end of a subpass so you would always fail validation if a pipeline in the latter subpass used fewer inputs.
2. Validation aspects are now reset when the pipeline is set so that we re-check if all required inputs are required.
3. @kainino0x noticed that for some reason `EnsureTextureUsage` checks if the texture is frozen as an output attachment


src/backend/CommandBufferStateTracker.cpp
```diff
@@ -542,10 +549,20 @@ namespace backend {
...
+            1 << VALIDATION_ASPECT_VERTEX_BUFFERS |
+            1 << VALIDATION_ASPECT_INDEX_BUFFER;
+        aspects &= ~pipelineDependentAspects;
+    }
+
     bool CommandBufferStateTracker::SetPipelineCommon(PipelineBase* pipeline) {
         PipelineLayoutBase* layout = pipeline->GetLayout();
 
-        aspects.reset(VALIDATION_ASPECT_BIND_GROUPS);
+        ResetDependentPipelineApsects();
```
Kangz: All we want to reset here are the lazily recomputed aspects. Would it be enough to add the ```inputsSet.reset()``` in EndSubPass and ```aspects.reset(VALIDATION_ASPECT_VERTEX_BUFFERS);``` here?

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -542,10 +549,20 @@ namespace backend {
         return true;
     }
 
+    void CommandBufferStateTracker::ResetDependentPipelineApsects() {
```
kainino0x: nit: `DependentPipeline`->`PipelineDependent` and `Apsects`->`Aspects` if this function is kept

src/backend/CommandBufferStateTracker.cpp
kainino0x: I think `VALIDATION_ASPECT_BIND_GROUPS` should get reset here too, but those two may be enough.

Kangz: Yeah VERTEX_BUFFERS would be in addition to the BIND_GROUPS currently present in the code.

austinEng: Updated this

kainino0x: oh, yeah, I was looking at the split view so I forgot that was there before.

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -289,9 +290,12 @@ namespace backend {
             builder->HandleError("Can't use a compute pipeline while a render pass is active");
             return false;
         }
-        aspects.set(VALIDATION_ASPECT_COMPUTE_PIPELINE);
+        if (!SetPipelineCommon(pipeline)) {
```
Kangz: Now we could revert this to ```return SetPipelineCommon(pipeline)``` like it did before, as the ordering doesn't matter anymore. (sorry for all the back and forths.)

austinEng:
@kainino0x Added a couple tests

src/tests/unittests/validation/RenderCommandsValidationTests.cpp
```diff
@@ -0,0 +1,258 @@
+// Copyright 2017 The NXT Authors
```
Kangz: Can this be in a VertexBufferValidationTests.cpp file and have more stuff factored (shader creation, creation of a bunch of vertex buffers, etc) so we can easily update it if the API changes, and easily add more tests?

PR #116, opened 2017-08-14
Allow CMake to find next.json if NXT is built by another project

austinEng:
Allows CMake to find `next.json` if NXT is built by another project


Kangz:
LGTM too.

PR #118, opened 2017-08-15
Configure texture usage for Swap Chains

austinEng:
Right now configuring initial / allowed usage only affects the D3D12 binding. Not sure if anything needs to change for Metal.

### DIFF
--- a/examples/Animometer.cpp
+++ b/examples/Animometer.cpp
@@ -49,7 +49,7 @@ void init() {
 
     queue = device.CreateQueueBuilder().GetResult();
     swapchain = GetSwapChain(device);
-    swapchain.Configure(nxt::TextureFormat::R8G8B8A8Unorm, nxt::TextureUsageBit::OutputAttachment | nxt::TextureUsageBit::Present, nxt::TextureUsageBit::OutputAttachment, 640, 480);
+    swapchain.Configure(nxt::TextureFormat::R8G8B8A8Unorm, nxt::TextureUsageBit::OutputAttachment, nxt::TextureUsageBit::OutputAttachment, 640, 480);
 
     nxt::ShaderModule vsModule = utils::CreateShaderModule(device, nxt::ShaderStage::Vertex, R"(
         #version 450
--- a/examples/Animometer.cpp
+++ b/examples/Animometer.cpp
@@ -49,7 +49,7 @@ void init() {
 
     queue = device.CreateQueueBuilder().GetResult();
     swapchain = GetSwapChain(device);
-    swapchain.Configure(nxt::TextureFormat::R8G8B8A8Unorm, 640, 480);
+    swapchain.Configure(nxt::TextureFormat::R8G8B8A8Unorm, nxt::TextureUsageBit::OutputAttachment | nxt::TextureUsageBit::Present, nxt::TextureUsageBit::OutputAttachment, 640, 480);
 
     nxt::ShaderModule vsModule = utils::CreateShaderModule(device, nxt::ShaderStage::Vertex, R"(
         #version 450
--- a/examples/glTFViewer/glTFViewer.cpp
+++ b/examples/glTFViewer/glTFViewer.cpp
@@ -464,7 +464,7 @@ namespace {
 
         queue = device.CreateQueueBuilder().GetResult();
         swapchain = GetSwapChain(device);
-        swapchain.Configure(nxt::TextureFormat::R8G8B8A8Unorm, 640, 480);
+        swapchain.Configure(nxt::TextureFormat::R8G8B8A8Unorm, nxt::TextureUsageBit::OutputAttachment | nxt::TextureUsageBit::Present, nxt::TextureUsageBit::OutputAttachment, 640, 480);
 
         renderpass = CreateDefaultRenderPass(device);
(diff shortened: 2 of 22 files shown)


examples/Animometer.cpp
```diff
@@ -49,7 +49,7 @@ void init() {
 
     queue = device.CreateQueueBuilder().GetResult();
     swapchain = GetSwapChain(device);
-    swapchain.Configure(nxt::TextureFormat::R8G8B8A8Unorm, 640, 480);
```
kainino0x: Would it make sense to make the `Present` usage here implicit? It seems like it's going to be required no matter what.

kainino0x: er, not here, but in the new version.

austinEng: Seems fine to me

Kangz: +1

examples/glTFViewer/glTFViewer.cpp
```diff
@@ -464,7 +464,7 @@ namespace {
 
         queue = device.CreateQueueBuilder().GetResult();
         swapchain = GetSwapChain(device);
-        swapchain.Configure(nxt::TextureFormat::R8G8B8A8Unorm, 640, 480);
+        swapchain.Configure(nxt::TextureFormat::R8G8B8A8Unorm, nxt::TextureUsageBit::OutputAttachment | nxt::TextureUsageBit::Present, nxt::TextureUsageBit::OutputAttachment, 640, 480);
```
Kangz: Stray present usage bit here.

src/backend/SwapChain.cpp
```diff
@@ -34,17 +34,23 @@ namespace backend {
         return device;
     }
 
-    void SwapChainBase::Configure(nxt::TextureFormat format, uint32_t width, uint32_t height) {
+    void SwapChainBase::Configure(nxt::TextureFormat format, nxt::TextureUsageBit allowedUsage, nxt::TextureUsageBit initialUsage, uint32_t width, uint32_t height) {
         if (width == 0 || height == 0) {
             device->HandleError("Swap chain cannot be configured to zero size");
             return;
         }
+        if ((initialUsage & allowedUsage) == nxt::TextureUsageBit::None) {
```
Kangz: this should be (initialUsage & allowedUsage) != initialUsage

src/backend/SwapChain.cpp
```diff
@@ -59,8 +65,8 @@ namespace backend {
         builder->SetExtent(width, height, 1);
         builder->SetFormat(format);
         builder->SetMipLevels(1);
-        builder->SetAllowedUsage(nxt::TextureUsageBit::OutputAttachment | nxt::TextureUsageBit::Present);
-        builder->SetInitialUsage(nxt::TextureUsageBit::OutputAttachment);
+        builder->SetAllowedUsage(allowedUsage);
```
Kangz: allowedUsage | nxt::TextureUsageBit::Present

next.json
```diff
@@ -1050,6 +1050,8 @@
                 "name": "configure",
                 "args": [
                     {"name": "format", "type": "texture format"},
+                    {"name": "allowedUsage", "type": "texture usage bit"},
+                    {"name": "initialUsage", "type": "texture usage bit"},
```
Kangz: @kainino0x: didn't we say that when the next texture is provided it should be in the None usage so we force a transition? This would remove the need for initialUsage.

src/backend/SwapChain.cpp
kainino0x: I think it should check that initialUsage has exactly one bit, making it:

Then this check becomes `if (!(HasZeroOrOneBits(initialUsage) && (initialUsage & allowedUsage)))` or something like that.

src/backend/SwapChain.cpp
kainino0x: This is already done in `this->allowedUsage =` above

next.json
kainino0x: We don't force an explicit transition, but we ended up making the transition implicit in `BeginRenderSubpass`. So we don't need `SwapChainBuilder::SetInitialUsage` for renderpass attachments, but it could be convenient for blits? Not a big deal either way.

austinEng: Does this guarantee the usage bit at the start of the subpass? I remember still needing to include an explicit transition.

Kangz:
Actually, now that I think of it, should we only allow pure write formats for swapchain textures so the application cannot see garbage? Or should we allow any format but have an implicit clear?

src/backend/SwapChain.cpp
Kangz: If we allow read usages, it would be ok to have ```initialUsage = nxt::TextureUsageBit::TransferSrc | nxt::TextureUsageBit::Sampled```

src/backend/SwapChain.cpp
Kangz: thanks

austinEng:
Well, if we want to avoid seeing garbage we'll have to do a clear. Present is still a read format so if the application does absolutely nothing, you'll see garbage unless you clear.

kainino0x:
Closing this and pushing it to a new PR; see #123

Kangz:
Thanks for picking this up!

PR #119, opened 2017-08-17
Validate 256-byte alignment on buffer view offset for bind groups

austinEng:


src/common/Math.h
```diff
@@ -23,8 +23,9 @@ uint32_t ScanForward(uint32_t bits);
 uint32_t Log2(uint32_t value);
 bool IsPowerOfTwo(size_t n);
 
-bool IsAligned(const void* ptr, size_t alignment);
+bool IsAlignedPtr(const void* ptr, size_t alignment);
```
Kangz: nit: IsPtrAligned?

PR #121, opened 2017-08-18
Add push constant tests

Kangz:
Also adds a bunch of fixes for push constants (and others) on OpenGL and Metal.


src/tests/end2end/PushConstantTests.cpp
```diff
@@ -0,0 +1,459 @@
...
+        .GetResult();
+
+    queue.Submit(1, &commands);
+
+    EXPECT_BUFFER_U32_EQ(1, binding1.resultBuffer, 0);
+    EXPECT_BUFFER_U32_EQ(1, binding2.resultBuffer, 0);
+}
+
+// Try setting all push constants
+TEST_P(PushConstantTest, SetAllConstantsToNoneZero) {
```
kainino0x: `NonZero`*?

src/tests/end2end/PushConstantTests.cpp
```diff
@@ -0,0 +1,459 @@
...
+    PushConstantSpec spec;
+    std::array<uint32_t, kMaxPushConstants> values;
+    for (uint32_t i = 0; i < kMaxPushConstants; ++i) {
+        spec.push_back({Int, static_cast<int>(i)});
```
kainino0x: First one is zero, maybe `(i + 1)` if you want to avoid that?

src/tests/end2end/PushConstantTests.cpp
Kangz: Done

src/tests/end2end/PushConstantTests.cpp
Kangz: Done

Kangz:
Thanks for the review!

PR #123, opened 2017-08-29
Configure texture usage for Swap Chains

kainino0x:
Resurrection of #118

### DIFF
--- a/src/utils/D3D12Binding.cpp
+++ b/src/utils/D3D12Binding.cpp
@@ -77,6 +77,41 @@ namespace utils {
 
             return factory;
         }
+
+        DXGI_USAGE D3D12SwapChainBufferUsage(nxtTextureUsageBit allowedUsages) {
+            DXGI_USAGE usage = DXGI_CPU_ACCESS_NONE;
+            if (allowedUsages & NXT_TEXTURE_USAGE_BIT_SAMPLED) {
+                usage |= DXGI_USAGE_SHADER_INPUT;
+            }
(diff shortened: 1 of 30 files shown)


kainino0x:
I'll add the last few comments from the original review soon and send this for review.

Not sure how to fix the CLA issue. I guess the easiest is to either `--reset-author` or admin-force-merge.

Kangz:
I'd rather not admin-force merge, Austin's contribution was done while at Google so we can reset-author without issues.

kainino0x:
Answering remaining questions on the original PR.

> @kainino0x: didn't we say that when the next texture is provided it should be in the None usage so we force a transition? This would remove the need for initialUsage.

I removed initialUsage in my last commit but I'm not set on that decision. @Kangz WDYT?

> The D3D12Binding still does a transition when it swaps buffers. Should this be removed/moved to BeginRenderSubpass?

I removed it.

> Actually, now that I think of it, should we only allow pure write formats for swapchain textures so the application cannot see garbage? Or should we allow any format but have an implicit clear?

We decided to have an implicit clear (but I have not inserted it).

kainino0x:
Hey @Kangz, I realized this is still floating around. PTAL.

src/utils/D3D12Binding.cpp
```diff
@@ -77,6 +77,41 @@ namespace utils {
 
             return factory;
         }
+
+        DXGI_USAGE D3D12SwapChainBufferUsage(nxtTextureUsageBit allowedUsages) {
```
Kangz: Eventually we'll need to find a way to factor this. No need to address now.

PR #133, opened 2017-11-15
Index in pipeline

Kangz:
PTAL, this is an old-ish branch that I finally finished tidying.

### DIFF
--- a/src/tests/end2end/IndexFormatTests.cpp
+++ b/src/tests/end2end/IndexFormatTests.cpp
@@ -1,0 +1,268 @@
+// Copyright 2017 The NXT Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "tests/NXTTest.h"
+
+#include "common/Assert.h"
+#include "utils/NXTHelpers.h"
+
+constexpr uint32_t kRTSize = 400;
+
+class IndexFormatTest : public NXTTest {
+    protected:
+        void SetUp() override {
+            NXTTest::SetUp();
+
+            renderpass = device.CreateRenderPassBuilder()
+                .SetAttachmentCount(1)
+                .AttachmentSetFormat(0, nxt::TextureFormat::R8G8B8A8Unorm)
+                .AttachmentSetColorLoadOp(0, nxt::LoadOp::Clear)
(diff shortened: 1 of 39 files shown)


src/tests/end2end/IndexFormatTests.cpp
```diff
@@ -0,0 +1,268 @@
...
+// Test that the Uint16 index format is correctly interpreted
+TEST_P(IndexFormatTest, Uint16) {
+    nxt::RenderPipeline pipeline = MakeTestPipeline(nxt::IndexFormat::Uint16);
+
+    nxt::Buffer vertexBuffer = utils::CreateFrozenBufferFromData<float>(device, nxt::BufferUsageBit::Vertex, {
+        -1.0f,  1.0f, 0.0f, 1.0f,
+         1.0f,  1.0f, 0.0f, 1.0f,
+        -1.0f, -1.0f, 0.0f, 1.0f
+    });
+    // If this is interpreted as uint32, it will have index 1 and 2 be both 0 and render nothing
```
kainino0x: The first index would be 0x00010002 or something, so wouldn't it potentially throw an error, failing to test the actual problem? (Although if the error got skipped because validation was done for Uint16, and then the driver received Uint32, then I imagine we'd either get garbage/crash or no rendering, so it's probably fine.)

src/tests/end2end/IndexFormatTests.cpp
```diff
@@ -0,0 +1,268 @@
...
+    nxt::Buffer indexBuffer = utils::CreateFrozenBufferFromData<uint16_t>(device, nxt::BufferUsageBit::Index, {
+        1, 2, 0, 0 ,0 ,0
```
kainino0x: nit: spaces around commas

src/tests/end2end/IndexFormatTests.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+// Test for primitive restart use vertices like in the drawing and draw the following
+// indices: 0 1 2 PRIM_RESTART 3 4 2. Then A and B should be written but not C.
+//             0
+//             |\
+//             |B \
+//             2---1
+//            /| C
+//          / A|
+//         4---3
```
kainino0x: If, say, primitive restart were disabled, wouldn't this again go out of bounds?

If the restart index has to be 0xFFFFFFFFu, this isn't easily solvable for the Uint32 case (I think), but the Uint16 case could easily have a large enough vertex buffer that 0xFFFFu is a valid index.

src/tests/end2end/IndexFormatTests.cpp
Kangz: I'm hoping we get a crash there. I agree the tests aren't perfect but the other options would be to go full dEQP.

src/tests/end2end/IndexFormatTests.cpp
Kangz: It would crash hopefully.

src/tests/end2end/IndexFormatTests.cpp
Kangz: Done

Kangz:
Thanks for the review.

PR #134, opened 2017-11-20
Vulkan up to instance creation

Kangz:
PTAL


src/backend/vulkan/VulkanBackend.cpp
```diff
@@ -103,6 +160,84 @@ namespace vulkan {
...
+                layersToRequest.push_back(kLayerNameLunargStandardValidation);
+                usedGlobals->standardValidation = true;
+            }
+            if (info.global.debugReport) {
+                extensionsToRequest.push_back(kExtensionNameExtDebugReport);
+                usedGlobals->debugReport = true;
+            }
+        #endif
+
+        VkApplicationInfo appInfo;
```
kainino0x: If you `= {};` here, you could probably omit the `= nullptr` and `= 0`s below? If you want.

Kangz: I'd rather not omit such ` = 0`. Though the `pNext = nullptr` is annoying to write. Keeping as is if you don't mind.

Kangz:
Thanks for the review! Landing now.

PR #135, opened 2017-11-20
Lazy vertex/index buffer application in GL

Kangz:
PTAL

### DIFF
--- a/src/backend/opengl/CommandBufferGL.cpp
+++ b/src/backend/opengl/CommandBufferGL.cpp
@@ -17,6 +17,7 @@
 #include "backend/Commands.h"
 #include "backend/opengl/BufferGL.h"
 #include "backend/opengl/ComputePipelineGL.h"
+#include "backend/opengl/InputStateGL.h"
 #include "backend/opengl/OpenGLBackend.h"
 #include "backend/opengl/PersistentPipelineStateGL.h"
 #include "backend/opengl/PipelineLayoutGL.h"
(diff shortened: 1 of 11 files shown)


src/backend/opengl/InputStateGL.cpp
```diff
@@ -0,0 +1,65 @@
...
+        glBindVertexArray(vertexArrayObject);
+        auto& attributesSetMask = GetAttributesSetMask();
+        for (uint32_t location = 0; location < attributesSetMask.size(); ++location) {
+            if (!attributesSetMask[location]) {
+                continue;
+            }
+            auto attribute = GetAttribute(location);
+            glEnableVertexAttribArray(location);
+
+            attributesUsingInput[attribute.bindingSlot][location] = true;
```
Kangz: This line is the only change in this file.

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -85,9 +109,87 @@ namespace opengl {
...
+        };
+
+        // Vertex buffers and index buffers are implemented as part of an OpenGL VAO that corresponds to an
+        // InputState. On the contrary in NXT they are part of the global state. This means that we have to
+        // re-apply these buffers on an InputState change.
+        struct InputBufferTracker {
+            bool indexBufferDirty = false;
+            Buffer* indexBuffer = nullptr;
+
+            std::bitset<kMaxVertexAttributes> dirtyVertexBuffers;
```
kainino0x: Why `kMaxVertexAttributes` here and `kMaxVertexInputs` below? Seems like they're both indexed the same way in `OnSetVertexBuffers`.

Kangz: Whoops, code was wrong. Done.

Kangz:
Thanks for the review!

PR #137, opened 2017-11-22
Vulkan: hack SetSubData and MapRead

Kangz:
PTAL, at this point the VulkanBackend can pass the SetSubData tests with the NXTTest swapchain related code commented out. This is a hack however because everything stays inside `vulkan::Buffer`. I'll add a proper queue, GPU copies, and fencing next. Then a proper "buffer uploader" for SetSubData.


Kangz:
Thanks for the review!

PR #138, opened 2017-11-22
Get to first passing test on Vulkan

Kangz:
PTAL


Kangz:
@kainino0x you might be doing this already, but looking at commits in order should be simpler than looking at the diff for the whole PR.

PR #139, opened 2017-11-23
Rename member variables from camelCase to mCamelCase

Kangz:
TBR


Kangz:
Marked TBR, but it feel free to skip it.

PR #140, opened 2017-11-23
Build the Vulkan backend on CI

Kangz:


Kangz:
PTAL, this should pass CI now.

PR #141, opened 2017-11-27
Format src/ except src/test

Kangz:
WIP: intended to fail with the clang-format travis hook


Kangz:
PTAL, this one should pass tests. I also checked that the check for the formatting works correctly.

.travis.yml
```diff
@@ -20,8 +20,11 @@ addons:
  # Everything under apt is only used on linux
  apt:
   sources:
-   # PPAs to get newer toolchains
+   # PPA for clang 5.0
    - llvm-toolchain-trusty-5.0
+   # PPA for clang-fromat 6.0
```
kainino0x: "fromat"

Kangz: Done

PR #142, opened 2017-11-27
Travis: use trusty distro

Kangz:


kainino0x:
For reference, the previous attempt at this was #111

Kangz:
Thanks. I think what held us back was the lack of whitelisted packages that looks ok now.

Kangz:
SUCCESS!

PR #144, opened 2017-12-01
Travis: install pip manually on OSX

Kangz:
PTAL, this will fix OSX CI.


Kangz:
Thanks. This passed one of the OSX tests so I'll submit it without waiting for the other. Travis has a HUGE backlog of OSX jobs right now.

PR #146, opened 2017-12-04
Implement buffer copies and transitions

Kangz:
PTAL


Kangz:
Thanks for the review. I will rebase this on top of the Vulkan CI once it is landed.

Kangz:
PTAL again, I added a commit at the beginning of the stack that should fix the compile error on 32bit.

kainino0x:
Needs reformat though

Kangz:
Merging manually, clang-format is having a formatting loop.

PR #152, opened 2018-01-16
Preliminary changes for the Vulkan SwapChain work.

Kangz:
PTAL. This is expect to fail CI because of formatting change ping-ponging in vulkan_platform.h


Kangz:
Thanks!

PR #153, opened 2018-01-26
Vulkan

Kangz:
PTAL, these are some other commits before the meat of the swapchain implementation in the Vulkan backend.


Kangz:
The compile error is legit, fixing it.

Kangz:
Finally passes Travis, thanks for the review!

PR #154, opened 2018-01-29
Fix Travis after LLVM APT repo change

Kangz:
After LLVM 6.0 was branched, the LLVM APT repositories changed so that
clang-format-6.0 isn't in the "latest" repo but in the "6.0" repo.
Update travis.yml for this, and start using clang-6.0 instead of 5.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ addons:
   sources:
    # PPA for clang 5.0
    - llvm-toolchain-trusty-5.0
-   # PPA for clang-format 6.0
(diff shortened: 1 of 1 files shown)


Kangz:
I tried using the LLVM-6.0 repo but it isn't present in the Travis repo whitelist yet, so I'll change things to use clang-format 7.0 instead.

Made a [request](https://github.com/travis-ci/apt-source-whitelist/issues/378) to add LLVM 6.0 to the whitelist.

Kangz:
Thanks!

PR #156, opened 2018-02-02
Vulkan: Implement BlendState, BindGroups and  UBOs

Kangz:
PTAL! I thought I could enable BlendStateTest just by implementing BlendState but it turns out it uses a UBO so I implemented that too!


kainino0x:
Awesome, great progress!

Kangz:
Thanks!

PR #157, opened 2018-02-05
Vulkan: Enable DepthStencilStateTests

Kangz:
PTAL!

### DIFF
--- a/src/backend/vulkan/CommandBufferVk.cpp
+++ b/src/backend/vulkan/CommandBufferVk.cpp
@@ -269,6 +269,12 @@ namespace backend { namespace vulkan {
                     lastRenderPipeline = pipeline;
                 } break;
 
+                case Command::SetStencilReference: {
+                    SetStencilReferenceCmd* cmd = mCommands.NextCommand<SetStencilReferenceCmd>();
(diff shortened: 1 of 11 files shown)


src/backend/vulkan/CommandBufferVk.cpp
```diff
@@ -165,8 +165,6 @@ namespace backend { namespace vulkan {
                     device->fn.CmdSetLineWidth(commands, 1.0f);
                     device->fn.CmdSetDepthBounds(commands, 0.0f, 1.0f);
 
-                    device->fn.CmdSetStencilCompareMask(commands, VK_STENCIL_FRONT_AND_BACK, 0);
-                    device->fn.CmdSetStencilWriteMask(commands, VK_STENCIL_FRONT_AND_BACK, 0);
```
kainino0x: Removing this seems like it removes the default values for stencil compare/write masks. Is that okay?

Kangz: These parameters are now part of the pipeline state (and forced to be there by D3D12 I believe). Since every pipeline will set its own masks, there is no need to have default values anymore.

PR #158, opened 2018-02-06
D3D12: Fix swapchain, make it have a "NativeSwapChainImpl"

Kangz:


src/backend/d3d12/NativeSwapChainImplD3D12.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+        mBufferSerials[mCurrentBuffer] = mDevice->GetSerial();
+        return NXT_SWAP_CHAIN_NO_ERROR;
+    }
+
+    nxt::TextureFormat NativeSwapChainImpl::GetPreferredFormat() const {
+        return nxt::TextureFormat::R8G8B8A8Unorm;
+    }
+
+}}  // namespace backend::d3d12
```
kainino0x: nit: missing newline, apparently

Kangz: weird, I would have though clang-format would catch that. Done.

PR #160, opened 2018-02-07
Reduce AppVeyor MSBuild verbosity

Kangz:


Kangz:
Doesn't work :(

msiglreith:
maybe slightly uglier but `cmake --build . -- /verbosity:minimal` should also work

PR #161, opened 2018-02-08
dx12: Handle present texture in a specially

msiglreith:
PRESENT is an exclusive flag in NXT and can't be combined with other flags. The existing implementation treats the D3D12_RESOURCE_STATE_PRESENT as flag which is confusing due to being 0.


Kangz:
Thanks for the PR, this makes a lot of sense and we should do the same change throughout the codebase for Present and OutputAttachment (as well as Storage for buffers). I'll do it after this lands.

Also can you look into signing the Google CLA? If you're employed by Mozilla you should be able to ask the Point of Contact to add you to the list. Otherwise you should fill it as an individual contributor.

msiglreith:
Thanks, I signed as individual contributor

Kangz:
Thanks!

PR #167, opened 2018-03-02
D3D12: Keep a reference to pipelines until unused

Kangz:
Previously we would remove the reference to pipelines in the destructor
of the d3d12::*Pipeline objects which could cause the D3D12 pipeline
state to be destroyed while still used by in-flight commands. Add a
global queue of ComPtrs to keep alive in the d3d12::Device to fix this.


Kangz:
Thanks you for the review!

PR #168, opened 2018-04-03
Address modes

SenorBlanco:
Implement sampler address modes (repeat, mirrored repeat, clamp-to-edge).
Add an end-to-end test for Samplers.

### DIFF
--- a/next.json
+++ b/next.json
@@ -714,6 +714,14 @@
             {"value": 1, "name":"linear"}
         ]
     },
+    "address mode": {
+        "category": "enum",
+        "values": [
+            {"value": 0, "name":"repeat"},
+            {"value": 1, "name":"mirrored repeat"},
+            {"value": 2, "name":"clamp to edge"}
+        ]
+    },
     "float": {
         "category": "native"
     },
@@ -1017,6 +1025,14 @@
                     {"name": "mag filter", "type": "filter mode"},
                     {"name": "min filter", "type": "filter mode"},
                     {"name": "mipmap filter", "type": "filter mode"}
+                ]
+            },
+            {
+                "name": "set address mode",
+                "args": [
+                    {"name": "address mode u", "type": "address mode"},
+                    {"name": "address mode v", "type": "address mode"},
+                    {"name": "address mode w", "type": "address mode"}
                 ]
             }
         ]
--- a/src/tests/end2end/SamplerTests.cpp
+++ b/src/tests/end2end/SamplerTests.cpp
@@ -1,0 +1,171 @@
+// Copyright 2018 The NXT Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <array>
+#include <cmath>
+
+#include "tests/NXTTest.h"
+
+#include "common/Assert.h"
+#include "common/Constants.h"
+#include "utils/NXTHelpers.h"
+
+constexpr static unsigned int kRTSize = 64;
+
+namespace {
+    struct AddressModeTestCase {
+        nxt::AddressMode mMode;
+        uint8_t mExpected2;
+        uint8_t mExpected3;
+    };
+    AddressModeTestCase addressModes[] = {
+        { nxt::AddressMode::Repeat,           0, 255, },
+        { nxt::AddressMode::MirroredRepeat, 255,   0, },
+        { nxt::AddressMode::ClampToEdge,    255, 255, },
+    };
+}
+
+class SamplerTest : public NXTTest {
+protected:
+    void SetUp() override {
+        NXTTest::SetUp();
+        mFB = utils::CreateBasicFramebuffer(device, kRTSize, kRTSize);
+        mFB.color.TransitionUsage(nxt::TextureUsageBit::OutputAttachment);
+
+        mBindGroupLayout = device.CreateBindGroupLayoutBuilder()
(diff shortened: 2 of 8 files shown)


src/tests/end2end/SamplerTests.cpp
```diff
@@ -0,0 +1,195 @@
...
+            .GetResult();
+
+        auto pipelineLayout = device.CreatePipelineLayoutBuilder()
+            .SetBindGroupLayout(0, mBindGroupLayout)
+            .GetResult();
+
+        auto vsModule = utils::CreateShaderModule(device, nxt::ShaderStage::Vertex, R"(
+            #version 450
+            layout(location = 0) in vec4 pos;
+            void main() {
```
Kangz: nit: to make the test even more minimal, you could have the vertices defined in the shader, and do dead code removal. See https://github.com/Kangz/nxt-standalone/blob/master/src/tests/end2end/BlendStateTests.cpp#L36

src/tests/end2end/SamplerTests.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+        auto texture = device.CreateTextureBuilder()
+            .SetDimension(nxt::TextureDimension::e2D)
+            .SetExtent(2, 2, 1)
+            .SetFormat(nxt::TextureFormat::R8G8B8A8Unorm)
+            .SetMipLevels(1)
+            .SetAllowedUsage(nxt::TextureUsageBit::TransferDst | nxt::TextureUsageBit::Sampled)
+            .GetResult();
+
+        RGBA8 data[64 * 2];
```
Kangz: There's a bunch of magic numbers here, can it be rephrased in term of kTextureRowPitchAlignment ? You should only need RGBA8 data[kTextureRowPitchAlignment / sizeof(RGBA8) + 2].

src/tests/end2end/SamplerTests.cpp
```diff
@@ -0,0 +1,195 @@
...
+        memset(data, 128, sizeof(data));
```
Kangz: nit: sizeof(data), or make data an std::array.

src/tests/end2end/SamplerTests.cpp
Kangz: nit: Also comment explaining what the created texture looks like.

src/tests/end2end/SamplerTests.cpp
```diff
@@ -0,0 +1,195 @@
...
+        EXPECT_PIXEL_RGBA8_EQ(expectedU2, mFB.color, 2, 0);
+        EXPECT_PIXEL_RGBA8_EQ(expectedU3, mFB.color, 3, 0);
+        EXPECT_PIXEL_RGBA8_EQ(expectedV2, mFB.color, 0, 2);
+        EXPECT_PIXEL_RGBA8_EQ(expectedV3, mFB.color, 0, 3);
+        // TODO: add tests for W address mode, once NXT supports 3D textures
+    }
+
+    // Test drawing a rect with a checkerboard texture with different address modes.
+    void DoAddressModeTest() {
+        for (auto u : addressModes) {
```
Kangz: This function's body could go in the TEST_P instead.

src/tests/end2end/SamplerTests.cpp
```diff
@@ -0,0 +1,195 @@
...
+    nxt::RenderPipeline mPipeline;
+    nxt::TextureView mTextureView;
+    nxt::Buffer mQuad;
+};
+
+TEST_P(SamplerTest, AddressMode) {
+    DoAddressModeTest();
+}
+
+NXT_INSTANTIATE_TEST(SamplerTest, D3D12Backend, MetalBackend, OpenGLBackend, VulkanBackend)
```
Kangz: This won't work on the Vulkan backend yet, can you remove it? Also did you check it work on D3D12, Metal and GL?

next.json
```diff
@@ -701,6 +701,14 @@
             {"value": 1, "name":"linear"}
         ]
     },
+    "address mode": {
```
Kangz: note: in WebGPU we would probably have a longer name like TextureAddressMode or SamplerAddressMode.

Kangz:
Sorry for the Travis failure, I'll take a look. It needs to be maintained once in a while because we use non-standard APT repos.

Kangz:
The Linux build was just a flake but it looks like a couple changes are required: https://travis-ci.org/google/nxt-standalone/jobs/361800404#L1103

src/tests/end2end/SamplerTests.cpp
SenorBlanco: Done.

src/tests/end2end/SamplerTests.cpp
SenorBlanco: It is sizeof(data). :)

The 128 was just to show data that shouldn't be copied/sampled. No need tho, removed.

src/tests/end2end/SamplerTests.cpp
SenorBlanco: Done.

SenorBlanco: Done.

src/tests/end2end/SamplerTests.cpp
SenorBlanco: Done.

next.json
SenorBlanco: Could do, but shouldn't FilterMode be TextureFilterMode, then?

SenorBlanco:
Didn't get those locally. Maybe my config is weird.

Fixed, anyway.

src/tests/end2end/SamplerTests.cpp
SenorBlanco: Removed.

Tested on all backends; passes.

next.json
Kangz: Or SamplerFilterMode, yeah let's not do that in NXT.

SenorBlanco:
Formatting fixed. AFAICT, the two remaining failures are Travis issues, unrelated to this change.

SenorBlanco:
Squashed to two commits (implementation, test).
Rebased to tip-of-tree.

Kangz:
Can we have one more rebase one this? Sorry for the hassle.

Kangz:
Thanks, landing now.

PR #169, opened 2018-04-03
Travis.yml: Fix pip and cache it on OSX

Kangz:


.travis.yml
```diff
@@ -9,13 +9,19 @@ sudo: false
 language:
  - cpp
 
+python:
```
Kangz: This is actually necessary :'-(

Kangz:
@kainino0x PTAL, this is the regular "travis osx broke again so fix it" commit.

Kangz:
Thanks!

PR #175, opened 2018-04-09
Fix for strict-aliasing issue.

SenorBlanco:
Use a memcpy() to avoid strict-aliasing warnings. Using a union instead would be ideal, but this works for now.

### DIFF
--- a/src/backend/opengl/CommandBufferGL.cpp
+++ b/src/backend/opengl/CommandBufferGL.cpp
@@ -129,8 +129,8 @@ namespace backend { namespace opengl {
                                 break;
                             case PushConstantType::Float:
                                 float value;
-                                // Use a memcpy to avoid strict-aliasing warnings, even if it is still
-                                // technically undefined behavior.
+                                // Use a memcpy to avoid strict-aliasing warnings, even if it is
+                                // still technically undefined behavior.
                                 memcpy(&value, &mValues[stage][constant], sizeof(value));
                                 glUniform1f(location, value);
                                 break;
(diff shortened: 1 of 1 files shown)


src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -125,8 +125,9 @@ namespace backend { namespace opengl {
                                              *reinterpret_cast<GLuint*>(&mValues[stage][constant]));
                                 break;
                             case PushConstantType::Float:
-                                glUniform1f(location,
-                                            *reinterpret_cast<GLfloat*>(&mValues[stage][constant]));
+                                glUniform1f(
```
Kangz: Could this be done with maybe a memcpy into a std::array<float, kMaxPushConstants>? It is still technically a UB but at least on a different line so we can comment on this being required to avoid a strict aliasing warning.

Kangz:
Thanks for the update, I changed the comment slightly.

kainino0x:
Wouldn't a localized change to use a union be quite simple?
```
union { float f; uint32_t u; } value;
value.u = mValues[stage][constant];
glUniform1f(location, value.f);
```
(But I'm not sure this is any less undefined?)

kenrussell:
We were just talking about this at lunch and apparently the only way to convert a float to integer bits is to use memcpy? No combination of casts or unions allows this to be done in a non-undefined fashion? Can this be true?

kainino0x:
The comment here says the memcpy is also undefined behavior, but I'm not sure why that's true.

Kangz:
I also talked about this with etienneb@, there is no valid way to interpret float bits to integers and vice-versa in C. IIUC technically floats could be stored in a separate memory, or have an 80bit representation.

Kangz:
Damn forgot to squash, here goes the pretty git history :)

PR #176, opened 2018-04-09
Change Buffer::SetSubData() start and offset to chars

SenorBlanco:
Change the uint32_t offsets in Buffer::SetSubData() to be uint8_ts, rather than uint32_ts.
This makes it practical to do subloads of 16-bit index buffers, for example.

### DIFF
--- a/examples/ComputeBoids.cpp
+++ b/examples/ComputeBoids.cpp
@@ -88,8 +88,8 @@ void initBuffers() {
             .GetResult();
 
         particleBuffers[i].SetSubData(0,
-            sizeof(Particle) * kNumParticles / sizeof(uint32_t),
-            reinterpret_cast<uint32_t*>(initialParticles.data()));
+            sizeof(Particle) * kNumParticles,
+            reinterpret_cast<uint8_t*>(initialParticles.data()));
     }
 }
 
--- a/src/tests/end2end/BufferTests.cpp
+++ b/src/tests/end2end/BufferTests.cpp
@@ -41,20 +41,20 @@ class BufferMapReadTests : public NXTTest {
         const void* mappedData = nullptr;
 };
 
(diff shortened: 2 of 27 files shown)


examples/ComputeBoids.cpp
```diff
@@ -88,7 +88,7 @@ void initBuffers() {
             .GetResult();
 
         particleBuffers[i].SetSubData(0,
-            sizeof(Particle) * kNumParticles / sizeof(uint32_t),
+            sizeof(Particle) * kNumParticles,
             reinterpret_cast<uint32_t*>(initialParticles.data()));
```
Kangz: Can we have the `const uint32_t*` changed to a `const uint8_t*` to be consistent with the size argument?

src/tests/end2end/BufferTests.cpp
```diff
@@ -209,7 +209,7 @@ TEST_P(BufferSetSubDataTests, SmallDataAtOffset) {
 
     constexpr uint32_t kOffset = 2000;
     uint32_t value = 298371;
-    buffer.SetSubData(kOffset / 4, 1, &value);
+    buffer.SetSubData(kOffset, sizeof(value), &value);
```
Kangz: Can we also have 1 byte setsubdata tests and at an offset that's not aligned?

examples/ComputeBoids.cpp
Kangz: This will also make the wire work correctly, right now it thinks size is in number of uint32_t

SenorBlanco: Done.

src/tests/end2end/BufferTests.cpp
SenorBlanco: Done.

PR #177, opened 2018-04-09
Implement recent vertex and texture formats on Vk.

SenorBlanco:
Implement unorm_rgba8 and unorm_rg8 vertex formats and
{uint8, unorm8} x {r, rg} texture formats on Vulkan backend.


Kangz:
LGTM

PR #180, opened 2018-04-10
Fix a DrawElements bug in the Metal backend.

SenorBlanco:
The indexOffset of the draw command was not being used. It must be
added to the indexBufferOffset, scaled by index component size.


SenorBlanco:
Note: the only end2end tests which exercise DrawElements() seem to be index format tests; let me know if there's another place I could add one.

src/backend/metal/CommandBufferMTL.mm
```diff
@@ -304,14 +304,15 @@ void EndSubpass() {
...
+                    size_t formatSize = IndexFormatSize(lastRenderPipeline->GetIndexFormat());
 
                     ASSERT(encoders.render);
                     [encoders.render
                         drawIndexedPrimitives:lastRenderPipeline->GetMTLPrimitiveTopology()
                                    indexCount:draw->indexCount
                                     indexType:lastRenderPipeline->GetMTLIndexType()
                                   indexBuffer:indexBuffer
-                            indexBufferOffset:indexBufferOffset
+                            indexBufferOffset:indexBufferOffset + draw->firstIndex * formatSize
```
kainino0x: This makes the name of the local variable `indexBufferOffset` slightly misleading. Maybe it could be `indexBufferBaseOffset` or something?

Kangz:
Yeah I'm not sure where to put a test either, we could have a DrawElementsTest file?

SenorBlanco:
OK, I'll work on a DrawElements Quality Program. ;)

SenorBlanco:
OK, added a basic DrawElements test, exercising zero and non-zero index offsets. Verified that the test fails with the fix reverted.

SenorBlanco:
Rebased.

PR #182, opened 2018-04-11
D3D12 fix for Framebuffer GPU refcount issue.

SenorBlanco:
Reference the outgoing heap ComPtr, so it doesn't get Released before
the GPU is done with it.

This fixes intermittent ComPtr assertions on the D3D12 backend.


Kangz:
I think the proper fix would be to use DescriptorHeapAllocator::Release in the proper places, and ASSERT that on destruction of the descriptor heap allocator, all heaps have been released.

The root of the problem is that the DescriptorHeapHandle [here](https://github.com/SenorBlanco/nxt-standalone/blob/30749d619930559e1e1292389aa8be85ca5e60fb/src/backend/d3d12/CommandBufferD3D12.cpp#L51) get overwritten but are never added to be tracked. We should make sure they are "released" after every submission so that they are tracked.

Kangz:
FYI I haven't forgotten about this, I'll make a patch that fixes the issue after the multi pass to single render pass change.

PR #183, opened 2018-04-12
 Expand allowable formats for color attachments

SenorBlanco:
Remove assert for command buffer color binding point texture format. It was triggering on BGRA8, R8, etc.

Here's my reasoning: all color formats we've added so far are renderable. I think only legacy formats (luminance, intensity) and compressed formats are non-renderable. And if we do add non-renderable formats, binding them should probably be caught at validation, not with an assert.


Kangz:
The assert is because we are doing float clears so the code wouldn't work with integer texture (which might not even be renderable, I'm not sure). Can we transform the assert to check that the texture is floating point? Also can you add a commit that fixes the error by checking an additional string [here](https://github.com/google/nxt-standalone/blob/master/scripts/travis_lint_format.sh#L27)?

Kangz:
Work around the warning LGTM, thanks!

PR #185, opened 2018-04-17
Vulkan fixes

Kangz:
PTAL, coming back to the Vulkan backend with updated validation layers uncovered a number of issues.


Kangz:
Thanks for the review!

PR #187, opened 2018-04-18
Vulkan: Implement compute.

Kangz:


src/backend/vulkan/CommandBufferVk.cpp
```diff
@@ -67,6 +67,54 @@ namespace backend { namespace vulkan {
...
+                // draw / dispatch.
+                if (mCurrentLayout == nullptr) {
+                    mCurrentLayout = layout;
+                }
+
+                // Bindgroups that are not inherited will be set again before any draw or
+                // dispatch. Resetting the bits also makes sure we don't have leftover dirty
+                // bindgroups that don't exist in the pipeline layout.
+                mDirtySets &= ~layout->InheritedGroupsMask(mCurrentLayout);
+                mCurrentLayout = layout;
```
kainino0x: This logic seems like it should be equivalent to
```cpp
if (mCurrentLayout) {
    mDirtySets &= ~layout->InheritedGroupsMask(mCurrentLayout);
} else {
    mDirtySets.reset();
}
mCurrentLayout = layout;
```
which seems clearer to me.

Separately, I'm a bit confused about the implementation of InheritedGroupsMask:
```cpp
std::bitset<kMaxBindGroups> PipelineLayoutBase::InheritedGroupsMask(
    const PipelineLayoutBase* other) const {
    return {GroupsInheritUpTo(other) - 1};
}
```
This doesn't look like a mask to me... maybe I'm missing something, but shouldn't it be something like `{(1 << GroupsInheritUpTo(other)) - 1}`?

Kangz: Done and done (good catch!)

PR #188, opened 2018-04-26
Fix some leaking builder objects in common backend.

SenorBlanco:
Delete the TextureBuilder created in SwapChainBase::GetNextTexture().
Delete the InputStateBuilder and DepthStencilStateBuilder created in
RenderPipelineBuilder::GetResultImpl().


SenorBlanco:
I wasn't sure if this was the best way to fix these. IWBN to use refcounted C++ objects, but these seem to be C objects.

Kangz:
Yes this is the correct approach. WebGPU looks like it won't have builders so this problem will disappear at that point.

PR #189, opened 2018-04-26
Fix a memory leak in metal ShaderModule.

SenorBlanco:


kainino0x:
Needs some formatting fixes, but LGTM

src/backend/metal/RenderPipelineMTL.mm
```diff
@@ -76,8 +76,8 @@ MTLIndexType MTLIndexFormat(nxt::IndexFormat format) {
             const auto& module = ToBackend(builder->GetStageInfo(stage).module);
 
             const auto& entryPoint = builder->GetStageInfo(stage).entryPoint;
-            id<MTLFunction> function =
-                module->GetFunction(entryPoint.c_str(), ToBackend(GetLayout())).function;
+            ShaderModule::MetalFunctionData data = module->GetFunction(entryPoint.c_str(), ToBackend(GetLayout()));
+            id<MTLFunction> function = data.function;
```
Kangz: Also ComputePipelineMTL will need a similar change.

src/backend/metal/RenderPipelineMTL.mm
```diff
@@ -76,8 +76,8 @@ MTLIndexType MTLIndexFormat(nxt::IndexFormat format) {
             const auto& module = ToBackend(builder->GetStageInfo(stage).module);
 
             const auto& entryPoint = builder->GetStageInfo(stage).entryPoint;
-            id<MTLFunction> function =
-                module->GetFunction(entryPoint.c_str(), ToBackend(GetLayout())).function;
+            ShaderModule::MetalFunctionData data = module->GetFunction(entryPoint.c_str(), ToBackend(GetLayout()));
```
Kangz: I don't think this is correct because `data` will go out of scope at the end of the loop. They should be stored outside the loop in a `PerStage<ShaderModule::MetalFunctionData>`. The longer term solution would be to have a smart pointer that does the ObjC retain and release.

SenorBlanco: According to the docs, the vertexFunction and fragmentFunction properties are declared with the "strong" attribute:

@property(readwrite, nonatomic, strong) id<MTLFunction> vertexFunction;

In non-ARC mode, this corresponds to "retain", which calls retain on its argument. So the descriptor holds a ref to the function. (If it didn't retain, and only released, this wouldn't be a memory leak -- it would just adopt the creation ref.)

So I think the change is correct as-is, but I'm no expert on this stuff.

Kangz: You look like an expert to me :) Thanks for the explanation, I think it makes sense as much as ObjC refcounting can be.

Kangz:
Merging, the proper fix of using ObjCRefPtrs will come later down the line.

PR #190, opened 2018-04-27
Fix some more memory leaks in the Metal backend.

SenorBlanco:


kainino0x:
This change is causing segfaults on my mac (got here via bisect). I'm looking into it.

Kangz:
This should already be fixed by https://github.com/google/nxt-standalone/commit/f77e3cc64800605cc36e4bba70d44481f908e0bc#diff-7e4fcfad0fb00f447d850554cc33d190

PR #191, opened 2018-04-27
Fix leak in Metal's Device::GetPendingCommandBuffer().

SenorBlanco:
GetPendingCommandBuffer() doesn't need to retain the metal
command buffer; it's created with a ref.


Kangz:
LGTM too, not sure why the clang format fails.

kainino0x:
I guess "no modified files to format" is an error case? There must be a way to avoid that.

SenorBlanco:
I tried to fix the lint error back in 6d90e01858da6a2084f047c21299b7d6c1c800ea, but it looks like I messed it up. Trying again.

Kangz:
Makes sense, LGTM thanks for fixing it.

PR #192, opened 2018-05-08
Pre-renderpass changes

Kangz:
@kainino0x, @SenorBlanco PTAL, these are changes required for before the PR that will change NXT to use "Metal style" render passes instead of "Vulkan style" ones.

### DIFF
--- a/src/backend/BindGroupLayout.cpp
+++ b/src/backend/BindGroupLayout.cpp
@@ -15,39 +15,19 @@
 #include "backend/BindGroupLayout.h"
 
 #include "backend/Device.h"
+#include "common/HashUtils.h"
 
 #include <functional>
 
 namespace backend {
 
     namespace {
-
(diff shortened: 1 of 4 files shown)


src/backend/metal/CommandBufferMTL.mm
```diff
@@ -144,7 +144,7 @@ void BeginSubpass(id<MTLCommandBuffer> commandBuffer, uint32_t subpass) {
                 }
 
                 render = [commandBuffer renderCommandEncoderWithDescriptor:descriptor];
-                [descriptor release];
```
SenorBlanco: As discussed, let's do these changes in a separate CL.

Kangz: Done

Kangz:
Thanks for the reviews, @SenorBlanco will put up a PR for the release fix.

PR #193, opened 2018-05-08
Don't release things which will be autoreleased.

SenorBlanco:


SenorBlanco:
Kai: I've updated skia-nxt to use an NSAutoreleasePool (as glfw does), and MacOS's "leaks" tool to monitor for leaks. No leaks are reported, and the autorelease pools's size is stable.

SenorBlanco:
Note: the headers do indicate that at least the command buffer returned by [MTLCommandQueue commandBuffer] is autoreleased:

@method commandBuffer
@abstract Returns a new autoreleased command buffer used to encode work into this queue that 
 maintains strong references to resources used within the command buffer.

Sadly, the others do not seem to be documented as autoreleased, so we have to rely on empirical behavior: if you use an NSAutoreleasePool, and try to explicitly release them, the app crashes. Without them, no leaks are observed.

kainino0x:
@SenorBlanco aha, I see. Sounds great!

PR #194, opened 2018-05-09
Simplify render passes

Kangz:
PTAL @kainino0x and @SenorBlanco


Kangz:
Whoops I just realized that the Metal changes got lost in between moving the branch to my different systems. I was able to get the changes back with a git reflog (phew ^^)

src/backend/RenderPassInfo.h
```diff
@@ -0,0 +1,109 @@
...
+    struct RenderPassDepthStencilAttachmentInfo {
+        nxt::LoadOp depthLoadOp;
+        nxt::LoadOp stencilLoadOp;
+        float clearDepth = 1.0f;
+        uint32_t clearStencil = 0;
+        Ref<TextureViewBase> view;
+    };
+
+    // RenderPassInfo contains the list of attachments for a renderpass along with data such as the
+    // load operation and the clear values for the attahcments.
```
SenorBlanco: Nit: attahcments -> attachments.

src/backend/metal/CommandBufferMTL.mm
```diff
@@ -129,11 +113,9 @@ void BeginSubpass(id<MTLCommandBuffer> commandBuffer, uint32_t subpass) {
                         descriptor.stencilAttachment.texture = texture;
                         descriptor.stencilAttachment.storeAction = MTLStoreActionStore;
 
-                        bool shouldClearStencilOnFirstUse =
```
SenorBlanco: I'm guessing this optimization is no longer applicable because subpasses are gone, so you always need to apply the loadop?

src/backend/vulkan/RenderPassCache.h
```diff
@@ -0,0 +1,81 @@
...
+
+        bool hasDepthStencil = false;
+        nxt::TextureFormat depthStencilFormat;
+        nxt::LoadOp depthLoadOp;
+        nxt::LoadOp stencilLoadOp;
+    };
+
+    // Caches VkRenderPasses so that we don't create duplicate ones for every RenderPipeline or
+    // render pass.
+    // TODO(cwallez@chromium.org): Make it an LRU cache somehow?
```
SenorBlanco: Out of curiosity, does this mean it grows without bound, currently?

src/tests/end2end/IndexFormatTests.cpp
```diff
@@ -24,36 +24,10 @@ class IndexFormatTest : public NXTTest {
...
-                .GetResult();
-
-            renderTargetView = renderTarget.CreateTextureViewBuilder().GetResult();
-
-            framebuffer = device.CreateFramebufferBuilder()
-                .SetRenderPass(renderpass)
-                .SetAttachment(0, renderTargetView)
-                .SetDimensions(kRTSize, kRTSize)
-                .GetResult();
+            renderPass = utils::CreateBasicRenderPass(device, kRTSize, kRTSize);
```
SenorBlanco: Nice refactor on these tests, BTW.

src/backend/RenderPassInfo.h
Kangz: Done

src/backend/metal/CommandBufferMTL.mm
Kangz: This was because the semantics of multi pass render passes were that attachments that need to be cleared are cleared at the beginning of the render pass so we cleared that on the first subpass they were used. This isn't needed anymore because there's only ever a single subpass that's the first one.

src/backend/vulkan/RenderPassCache.h
Kangz: Yes but given the small amount of real-world permutations that shouldn't cause problems before V1. For example in Skia's case the cache could grow to a whopping 10 elements!

src/tests/end2end/IndexFormatTests.cpp
Kangz: Thanks you :)

src/tests/unittests/validation/PushConstantsValidationTests.cpp
```diff
@@ -180,38 +151,32 @@ TEST_F(PushConstantTest, StageForComputePass) {
 }
 
 // Test valid stages for compute passes
```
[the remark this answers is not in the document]
Kangz: Done.

Kangz:
Thanks for the review, I'll wait for @kainino0x's before merging.

RenderPassInfo is pretty generic but in WebGPU it won't be a persistent object but just a "descriptor" that you pass on every BeginRenderPass. We just don't have a good way to do this in NXT currently so this gets us as close as WebGPU as possible without descriptor structures.

Kangz:
After offline discussion with @SenorBlanco, renamed RenderPassInfo to RenderPassDescriptor

src/backend/d3d12/RenderPassDescriptorD3D12.cpp
```diff
@@ -0,0 +1,78 @@
...
+    RenderPassDescriptor::RenderPassDescriptor(Device* device, RenderPassDescriptorBuilder* builder)
+        : RenderPassDescriptorBase(builder), mDevice(device) {
+        // Get and fill an RTV heap with the color attachments
+        uint32_t colorAttachmentCount = static_cast<uint32_t>(GetColorAttachmentMask().count());
+        if (colorAttachmentCount != 0) {
+            mRtvHeap = device->GetDescriptorHeapAllocator()->AllocateCPUHeap(
+                D3D12_DESCRIPTOR_HEAP_TYPE_RTV, colorAttachmentCount);
+        }
+
+        for (uint32_t i : IterateBitSet(GetColorAttachmentMask())) {
```
kainino0x: I think it would be clearer to move this `for` loop inside the preceding `if` block

src/backend/d3d12/RenderPassDescriptorD3D12.cpp
```diff
@@ -0,0 +1,78 @@
...
+            device->GetD3D12Device()->CreateDepthStencilView(resource.Get(), &dsvDesc, dsvHandle);
+        }
+    }
+
+    RenderPassDescriptor::OMSetRenderTargetArgs
+    RenderPassDescriptor::GetSubpassOMSetRenderTargetArgs() {
+        OMSetRenderTargetArgs args = {};
+
+        for (uint32_t i : IterateBitSet(GetColorAttachmentMask())) {
+            args.RTVs[i] = GetRTVDescriptor(i);
```
kainino0x: I think `RTVs[i]` is wrong here, shouldn't it have a separate counter (so that `RTVs` doesn't become sparse)?

src/backend/d3d12/FramebufferD3D12.cpp
```diff
@@ -1,95 +0,0 @@
...
-    }
-
-    Framebuffer::OMSetRenderTargetArgs Framebuffer::GetSubpassOMSetRenderTargetArgs(
-        uint32_t subpassIndex) {
-        const auto& subpassInfo = GetRenderPass()->GetSubpassInfo(subpassIndex);
-        OMSetRenderTargetArgs args = {};
-
-        for (uint32_t location : IterateBitSet(subpassInfo.colorAttachmentsSet)) {
-            uint32_t slot = subpassInfo.colorAttachments[location];
-            args.RTVs[args.numRTVs] = GetRTVDescriptor(slot);
```
kainino0x: cf. this

src/backend/metal/RenderPipelineMTL.mm
```diff
@@ -92,24 +92,16 @@ MTLIndexType MTLIndexFormat(nxt::IndexFormat format) {
...
-
-        if (subpassInfo.depthStencilAttachmentSet) {
-            const auto& attachmentInfo =
-                renderPass->GetAttachmentInfo(subpassInfo.depthStencilAttachment);
-            descriptor.depthAttachmentPixelFormat = MetalPixelFormat(attachmentInfo.format);
-            descriptor.stencilAttachmentPixelFormat = MetalPixelFormat(attachmentInfo.format);
+        if (HasDepthStencilAttachment()) {
+            nxt::TextureFormat depthStencilFormat = GetDepthStencilFormat();
+            descriptor.depthAttachmentPixelFormat = MetalPixelFormat(depthStencilFormat);
+            descriptor.stencilAttachmentPixelFormat = MetalPixelFormat(depthStencilFormat);
```
kainino0x: I know this was already here, but it looks like it will be wrong for depth-only or stencil-only formats (which I think we don't currently have). Could you add a TODO? (you can assign it to me)

next.json
```diff
@@ -879,70 +814,51 @@
...
+                    {"name": "clear b", "type": "float"},
+                    {"name": "clear a", "type": "float"}
                 ]
             },
             {
-                "name": "subpass set color attachment",
+                "name": "set depth stencil attachment",
                 "args": [
-                    {"name": "subpass index", "type": "uint32_t"},
-                    {"name": "output attachment location", "type": "uint32_t"},
```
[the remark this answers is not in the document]
Kangz: Not a problem, having named attachments in a small set [0, 4) is what D3D12 and Metal do.

src/backend/metal/RenderPipelineMTL.mm
Kangz: Done

src/backend/d3d12/RenderPassDescriptorD3D12.cpp
Kangz: Good point, especially since the heap size is the number of attachments. We need to figure out if we allow sparse attachments at some point. Done

Kangz:
Thanks for the thorough (half) review, I know this patch is massive and it is difficult to pick out functional changes from pure renaming.

src/tests/end2end/ScissorTests.cpp
```diff
@@ -98,86 +94,76 @@ TEST_P(ScissorTest, EmptyRect) {
         return;
     }
 
-    utils::BasicFramebuffer fb = utils::CreateBasicFramebuffer(device, 2, 2);
-    nxt::RenderPipeline pipeline = CreateQuadPipeline(fb.renderPass);
+    utils::BasicRenderPass renderPass = utils::CreateBasicRenderPass(device, 100, 100);
```
kainino0x: This test changed size, is that intentional?

Kangz: No, done

src/backend/vulkan/CommandBufferVk.cpp
```diff
@@ -184,85 +183,67 @@ namespace backend { namespace vulkan {
 
                 case Command::BeginRenderPass: {
                     BeginRenderPassCmd* cmd = mCommands.NextCommand<BeginRenderPassCmd>();
-                    Framebuffer* framebuffer = ToBackend(cmd->framebuffer.Get());
-                    RenderPass* renderPass = ToBackend(cmd->renderPass.Get());
+                    RenderPassDescriptor* info = ToBackend(cmd->info.Get());
 
                     // NXT has an implicit transition to color attachment on subpasses. Transition
```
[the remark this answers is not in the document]
Kangz: Done

PR #195, opened 2018-05-19
Sampler descriptor

Kangz:


generator/templates/backend/ValidationUtils.cpp
```diff
@@ -0,0 +1,40 @@
...
+//* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//* See the License for the specific language governing permissions and
+//* limitations under the License.
+
+#include "backend/ValidationUtils_autogen.h"
+
+namespace backend {
+
+    {% for type in by_category["enum"] %}
+        bool IsValid{{type.name.CamelCase()}}(nxt::{{as_cppType(type.name)}} value) {
```
kainino0x: Couldn't this be a "<" since our enums should be contiguous? Maybe with some extra asserts.

examples/HelloTriangle.cpp
```diff
@@ -56,9 +56,16 @@ void initTextures() {
         .SetAllowedUsage(nxt::TextureUsageBit::TransferDst | nxt::TextureUsageBit::Sampled)
         .GetResult();
 
-    sampler = device.CreateSamplerBuilder()
-        .SetFilterMode(nxt::FilterMode::Linear, nxt::FilterMode::Linear, nxt::FilterMode::Linear)
-        .GetResult();
+    {
```
kainino0x: Maybe not useful for this struct - but should we generate "fill with default" functions for struct initialization?

I think even constructors, like `nxt::SamplerDescriptor makeSamplerDescriptor()`, would work.

generator/templates/backend/ValidationUtils.cpp
Kangz: I was hoping that the switch would be future proof for when we start to have sparse enums for extensions. Compilers do a good job of optimizing this pattern when it is dense, see https://godbolt.org/g/HjWVZH
WDYT?

examples/HelloTriangle.cpp
Kangz: I thought about adding default to the generator but was worried that if WebGPU chooses different defaults and we update to follow them, we would have tests and samples break. Adding helper functions to make the default descriptor in NXTHelpers sounds good though. I'll look into it.

Kangz:
The only other changes could have been making end2end tests able to run on the wire, but it looks like it will be a different PR. What do you think of the overall direction for the structure of the backend/ code?

generator/templates/backend/ValidationUtils.cpp
kainino0x: SGTM.

kainino0x:
> What do you think of the overall direction for the structure of the backend/ code?

No problem with the changes in this commit. They seem pretty minor.

> Oh and I'll pushed for CI and wasn't looking for a review just yet (didn't do my own re-review). I'll look into having travis per-branch in my repo instead.

np, I can always review subsequent diffs if you don't squash.
I can also just not review things until you request review - so you don't need to set up travis again.

src/backend/d3d12/RenderPassDescriptorD3D12.cpp
```diff
@@ -57,7 +57,7 @@ namespace backend { namespace d3d12 {
     RenderPassDescriptor::GetSubpassOMSetRenderTargetArgs() {
         OMSetRenderTargetArgs args = {};
 
-        size_t rtvIndex = 0;
+        unsigned int rtvIndex = 0;
```
Kangz: This unrelated warning fix snuck in, but if that's ok I'll leave it in this commit.

kainino0x: "Leave the campground cleaner than you found it" :)

Kangz:
PTAL @SenorBlanco, also @kainino0x I PTAL at the latest commit.

Kangz:
Thank you, landing.

PR #199, opened 2018-06-11
WireCmd: factor code to [de]serialize structs and methods

Kangz:
Methods, when their arguments are stored in a structure, are very
similar to structures (duh?) Take advantage of this to factor the
[de]serialization code between them and not just memcpy structures
around. This will allow structures to contain objects, and pointers to
other structures / array of objects etc.


Kangz:
@SenorBlanco, @kainino0x PTAL.

I'll make a write-up of what these patches enable.

generator/templates/wire/WireCmd.cpp
```diff
@@ -14,8 +14,250 @@
...
+
+        //* const char* have their length embedded directly in the command.
+        {% for member in members if member.length == "strlen" %}
+            size_t {{as_varName(member.name)}}Strlen;
+        {% endfor %}
+    };
+
+    //* Returns the required transfer size for `record` in addition to the transfer structure.
+    size_t {{name}}GetExtraRequiredSize(const {{name}}& record) {
+        (void) record;
```
SenorBlanco: Nit: if this is to silence an unused param warning, IWBN to have an UNUSED macro just to make it clearer.

generator/templates/wire/WireCmd.cpp
```diff
@@ -14,8 +14,250 @@
...
+
+        size_t result = 0;
+
+        //* Special handling of const char* that have their length embedded directly in the command
+        {% for member in members if member.length == "strlen" %}
+            result += std::strlen(record.{{as_varName(member.name)}});
+        {% endfor %}
+
+        //* Gather how much space will be needed for pointer members.
+        {% for member in members if member.annotation != "value" and member.length != "strlen" %}
```
SenorBlanco: Nit: these tests (member.annotation != "value" and member.length != "strlen") appear a few times. Maybe refactor to a helper function that names what it is, rather than what it isn't? IsVariableSizedStruct or whatever?

generator/templates/wire/WireCmd.cpp
Kangz: Done

generator/templates/wire/WireCmd.cpp
Kangz: Agreed that the redundancy is annoying. Introducing functions in templates is a bit hard, because templates don't contain python code, just a very simple language that looks like python. Adding such a function would require changing `generator/main.py` or changing strlen char* pointer to not be pointers anymore (so we could iterate on members if member.annotation == "const*"). So I'd like to defer this.

SenorBlanco: SGTM. Was just a nit. :)

PR #204, opened 2018-06-15
Add Device::CreateQueue() instead of the builder

Kangz:
We are changing all object creation to use descriptors but there is no
creation argument to pass for queue, so instead Device::CreateQueue
takes no argument.


Kangz:
PTAL @kainino0x and @SenorBlanco

Kangz:
Thanks all!

PR #206, opened 2018-06-21
Use a descriptor for PipelineLayout

kainino0x:
Adds support for structures inside descriptors.

### DIFF
--- a/generator/main.py
+++ b/generator/main.py
@@ -153,8 +153,7 @@ def link_object(obj, types):
 
 def link_structure(struct, types):
     def make_member(m):
-        member = StructureMember(Name(m['name']), types[m['type']], m.get('annotation', 'value'))
-        return member
+        return StructureMember(Name(m['name']), types[m['type']], m.get('annotation', 'value'))
 
     members = []
     members_by_name = {}
--- a/generator/main.py
+++ b/generator/main.py
@@ -153,10 +153,21 @@ def link_object(obj, types):
 
 def link_structure(struct, types):
     def make_member(m):
-        return StructureMember(Name(m['name']), types[m['type']], m.get('annotation', 'value'))
+        member = StructureMember(Name(m['name']), types[m['type']], m.get('annotation', 'value'))
+        return member
 
-    # TODO(cwallez@chromium.org): Handle pointer members and their length
-    struct.members = [make_member(m) for m in struct.record['members']]
+    members = []
+    members_by_name = {}
+    for m in struct.record['members']:
+        member = make_member(m)
+        members.append(member)
+        members_by_name[member.name.canonical_case()] = member
+    struct.members = members
+
+    for (member, m) in zip(members, struct.record['members']):
+        # TODO(kainino@chromium.org): More robust pointer/length handling?
+        if 'length' in m:
+            member.length = members_by_name[m['length']]
 
 def parse_json(json):
     category_to_parser = {
--- a/generator/templates/wire/WireCmd.cpp
+++ b/generator/templates/wire/WireCmd.cpp
@@ -176,10 +176,10 @@
                 size_t memberLength = {{member_length(member, "record.")}};
                 auto memberBuffer = reinterpret_cast<{{member_transfer_type(member)}}*>(buffer);
 
+                buffer += memberLength * {{member_transfer_sizeof(member)}};
                 for (size_t i = 0; i < memberLength; ++i) {
                     {{serialize_member(member, "record." + memberName + "[i]", "memberBuffer[i]" )}}
                 }
-                buffer += memberLength * {{member_transfer_sizeof(member)}};
             }
         {% endfor %}
     }
--- a/src/backend/CommandBufferStateTracker.cpp
+++ b/src/backend/CommandBufferStateTracker.cpp
@@ -395,7 +395,8 @@ namespace backend {
         for (size_t i = 0; i < mBindgroups.size(); ++i) {
             if (auto* bindgroup = mBindgroups[i]) {
                 // TODO(kainino@chromium.org): bind group compatibility
-                if (bindgroup->GetLayout() != mLastPipeline->GetLayout()->GetBindGroupLayout(i)) {
+                auto* pipelineBGL = mLastPipeline->GetLayout()->GetBindGroupLayout(i);
+                if (pipelineBGL && bindgroup->GetLayout() != pipelineBGL) {
                     return false;
                 }
             }
--- a/src/backend/Device.cpp
+++ b/src/backend/Device.cpp
@@ -119,8 +119,21 @@ namespace backend {
     InputStateBuilder* DeviceBase::CreateInputStateBuilder() {
         return new InputStateBuilder(this);
     }
-    PipelineLayoutBuilder* DeviceBase::CreatePipelineLayoutBuilder() {
-        return new PipelineLayoutBuilder(this);
+    PipelineLayoutBase* DeviceBase::CreatePipelineLayout(const nxt::PipelineLayoutDescriptor* descriptor) {
+        MaybeError validation = ValidatePipelineLayoutDescriptor(this, descriptor);
+        if (validation.IsError()) {
+            // TODO(cwallez@chromium.org): Implement the WebGPU error handling mechanism.
+            delete validation.AcquireError();
+            return nullptr;
+        }
+
+        ResultOrError<PipelineLayoutBase*> maybePipelineLayout = CreatePipelineLayoutImpl(descriptor);
+        if (maybePipelineLayout.IsError()) {
+            // TODO(cwallez@chromium.org): Implement the WebGPU error handling mechanism.
+            delete validation.AcquireError();
+            return nullptr;
+        }
+        return maybePipelineLayout.AcquireSuccess();
     }
     QueueBase* DeviceBase::CreateQueue() {
         ResultOrError<QueueBase*> maybeQueue = CreateQueueImpl();
--- a/src/backend/Device.cpp
+++ b/src/backend/Device.cpp
@@ -119,7 +119,8 @@ namespace backend {
     InputStateBuilder* DeviceBase::CreateInputStateBuilder() {
         return new InputStateBuilder(this);
     }
-    PipelineLayoutBase* DeviceBase::CreatePipelineLayout(const nxt::PipelineLayoutDescriptor* descriptor) {
+    PipelineLayoutBase* DeviceBase::CreatePipelineLayout(
+        const nxt::PipelineLayoutDescriptor* descriptor) {
         MaybeError validation = ValidatePipelineLayoutDescriptor(this, descriptor);
         if (validation.IsError()) {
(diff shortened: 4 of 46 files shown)


kainino0x:
I think this is ready to review, but I don't think I'll be doing any more work this week so it's not urgent.

generator/main.py
```diff
@@ -153,10 +153,21 @@ def make_method(record):
 
 def link_structure(struct, types):
     def make_member(m):
-        return StructureMember(Name(m['name']), types[m['type']], m.get('annotation', 'value'))
-
-    # TODO(cwallez@chromium.org): Handle pointer members and their length
-    struct.members = [make_member(m) for m in struct.record['members']]
+        member = StructureMember(Name(m['name']), types[m['type']], m.get('annotation', 'value'))
```
Kangz: nit: member is only used by the return, so maybe we could leave the direct return?

generator/templates/wire/WireCmd.cpp
```diff
@@ -176,10 +176,10 @@
                 size_t memberLength = {{member_length(member, "record.")}};
                 auto memberBuffer = reinterpret_cast<{{member_transfer_type(member)}}*>(buffer);
 
+                buffer += memberLength * {{member_transfer_sizeof(member)}};
```
Kangz: Nice catch, this must have been tricky to find.

src/backend/Device.cpp
```diff
@@ -119,8 +119,21 @@ namespace backend {
...
+        if (validation.IsError()) {
+            // TODO(cwallez@chromium.org): Implement the WebGPU error handling mechanism.
+            delete validation.AcquireError();
+            return nullptr;
+        }
+
+        ResultOrError<PipelineLayoutBase*> maybePipelineLayout = CreatePipelineLayoutImpl(descriptor);
+        if (maybePipelineLayout.IsError()) {
+            // TODO(cwallez@chromium.org): Implement the WebGPU error handling mechanism.
+            delete validation.AcquireError();
```
Kangz: This should be `maybePipelineLayout.AcquireError()`

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -395,7 +395,8 @@ namespace backend {
         for (size_t i = 0; i < mBindgroups.size(); ++i) {
             if (auto* bindgroup = mBindgroups[i]) {
                 // TODO(kainino@chromium.org): bind group compatibility
-                if (bindgroup->GetLayout() != mLastPipeline->GetLayout()->GetBindGroupLayout(i)) {
+                auto* pipelineBGL = mLastPipeline->GetLayout()->GetBindGroupLayout(i);
+                if (pipelineBGL && bindgroup->GetLayout() != pipelineBGL) {
```
Kangz: This means it is ok to have extra bindgroups than just the ones that are in the pipeline layout. I think it is true.

src/backend/Error.h
```diff
@@ -36,7 +36,15 @@ namespace backend {
...
     //   NXT_RETURN_ERROR("My error message");
-#define NXT_RETURN_ERROR(EXPR) return MakeError(EXPR, __FILE__, __func__, __LINE__)
+#define NXT_RETURN_ERROR(MESSAGE) return MakeError(MESSAGE, __FILE__, __func__, __LINE__)
+#define NXT_TRY_ASSERT(EXPR, MESSAGE)                             \
+    {                                                             \
+        if (!(EXPR)) {                                            \
+            NXT_RETURN_ERROR(MESSAGE);                            \
+        }                                                         \
+    }                                                             \
+    for (;;)                                                      \
```
Kangz: Only `break` is required I think.

src/backend/PipelineLayout.cpp
```diff
@@ -20,10 +20,30 @@
...
-    PipelineLayoutBase::PipelineLayoutBase(PipelineLayoutBuilder* builder)
-        : mBindGroupLayouts(std::move(builder->mBindGroupLayouts)), mMask(builder->mMask) {
+    PipelineLayoutBase::PipelineLayoutBase(DeviceBase* device, const nxt::PipelineLayoutDescriptor* descriptor) {
+        ASSERT(descriptor->numBindGroupLayouts <= kMaxBindGroups);
+        for (uint32_t group = 0; group < descriptor->numBindGroupLayouts; ++group) {
+            mBindGroupLayouts[group] = reinterpret_cast<BindGroupLayoutBase*>(descriptor->bindGroupLayouts[group].Get());
+            mMask.set(group);
+        }
+        for (uint32_t group = descriptor->numBindGroupLayouts; group < kMaxBindGroups; ++group) {
+            auto builder = device->CreateBindGroupLayoutBuilder();
```
Kangz: Is this needed? Parts of the CL, for example `CommandBufferStateTracker.cpp` changes seem to assume no.

src/tests/end2end/PushConstantTests.cpp
```diff
@@ -49,9 +49,13 @@ class PushConstantTest: public NXTTest {
                 .SetBindingsType(kAllStages, nxt::BindingType::StorageBuffer, 0, extraBuffer ? 2 : 1)
                 .GetResult();
 
-            nxt::PipelineLayout pl = device.CreatePipelineLayoutBuilder()
-                .SetBindGroupLayout(0, bgl)
-                .GetResult();
+            nxt::PipelineLayout pl;
```
Kangz: I think this can use MakeBasicPipelineLayout too.

src/tests/unittests/WireTests.cpp
```diff
@@ -408,6 +408,39 @@ TEST_F(WireTests, StructureOfValuesArgument) {
     FlushClient();
 }
 
+// Test that the wire is able to send structures that contain objects
+TEST_F(WireTests, StructureOfObjectsArgument) {
+    nxtBindGroupLayoutBuilder bglBuilder = nxtDeviceCreateBindGroupLayoutBuilder(device);
+    nxtBindGroupLayoutBuilderSetBindingsType(bglBuilder, NXT_SHADER_STAGE_BIT_FRAGMENT, NXT_BINDING_TYPE_SAMPLER, 0, 1);
```
Kangz: nit: for the purpose of this test, I don't think we need to call `BindGroupLayoutBuilder::SetBindingsType`: empty layouts would be ok.

src/tests/unittests/WireTests.cpp
```diff
@@ -408,6 +408,39 @@ TEST_F(WireTests, StructureOfValuesArgument) {
...
+    descriptor.nextInChain = nullptr;
+    descriptor.numBindGroupLayouts = 1;
+    descriptor.bindGroupLayouts = &bgl;
+
+    nxtDeviceCreatePipelineLayout(device, &descriptor);
+    EXPECT_CALL(api, DeviceCreatePipelineLayout(apiDevice, MatchesLambda([apiBgl](const nxtPipelineLayoutDescriptor* desc) -> bool {
+        return desc->nextInChain == nullptr &&
+            desc->numBindGroupLayouts == 1 &&
+            desc->bindGroupLayouts[0] == apiBgl;
+            // TODO: do I need to somehow check that bgl's contents are correct?
```
Kangz: Nope this is already done by checking for `apiBgl` which is exactly the pointer that was returned for the BGL.

src/tests/unittests/WireTests.cpp
```diff
@@ -408,6 +408,39 @@ TEST_F(WireTests, StructureOfValuesArgument) {
     FlushClient();
 }
 
+// Test that the wire is able to send structures that contain objects
+TEST_F(WireTests, StructureOfObjectsArgument) {
```
Kangz: Thank you for adding this test! nit: I think it could be names "StructureOfObjectArrayArgument"

src/backend/vulkan/PipelineLayoutVk.cpp
```diff
@@ -22,8 +22,8 @@
 
 namespace backend { namespace vulkan {
 
-    PipelineLayout::PipelineLayout(PipelineLayoutBuilder* builder)
-        : PipelineLayoutBase(builder), mDevice(ToBackend(builder->GetDevice())) {
+    PipelineLayout::PipelineLayout(Device* device, const nxt::PipelineLayoutDescriptor* descriptor)
+        : PipelineLayoutBase(device, descriptor), mDevice(device) {
```
SenorBlanco: Since PipelineLayout only seems to need the FencedDeleter from the Device, I wonder if it should grab that here and stash it in a member instead. Just a thought.

generator/main.py
kainino0x: oops, must have happened when I was changing things.

src/tests/unittests/WireTests.cpp
kainino0x: Ah, this was extra from trying to catch the bug. But I guess it wasn't necessary after all.

src/tests/unittests/WireTests.cpp
kainino0x: The reason I still had this comment here is that the test didn't catch the bug. But apparently now it does.

src/tests/end2end/PushConstantTests.cpp
kainino0x: oops, missed this one when factoring it out, thanks.

src/backend/vulkan/PipelineLayoutVk.cpp
kainino0x: Not sure. It seems slightly simpler to keep it this way - it's more like other classes that save an mDevice, and it's simpler in case in the future we need something else from the Device.

src/backend/PipelineLayout.cpp
kainino0x: There is still some work left to be able to remove this. I'll leave a TODO to fix it.

generator/templates/wire/WireCmd.cpp
kainino0x: Took some time with a debugger (and noticing that `transfer` and `buffer` ended up being equal at some point). Unfortunately the test I wrote wasn't catching it (although now it is? confused about that.)

src/backend/Error.h
kainino0x: As opposed to `for (;;) break`? Doesn't seem to work. I copied this from NXT_TRY.

src/backend/vulkan/PipelineLayoutVk.cpp
SenorBlanco: OK, that's cool.

SenorBlanco:
Sounds good to me. But don't you need Kangz's approval too?

kainino0x:
Yes, but I meant since he is out, do you think it would be ok to land since I made the necessary changes? Or should I just wait (it's not a huge deal to wait)?

SenorBlanco:
Unless you're really backed/blocked on this, let's wait.

kainino0x:
Ok, I'll probably do another descriptor patch as a separate PR "on top of" this one so I won't have to deal with merging later.

PR #207, opened 2018-06-21
Assert HandleCommands succeeds

kainino0x:
This isn't ready to land yet because it causes at least one test to fail:

- [x] WireBufferMappingTests.UnmapCalledTooEarlyForRead

But @Kangz maybe you can take care of it if this is a problem.

### DIFF
--- a/examples/SampleUtils.cpp
+++ b/examples/SampleUtils.cpp
@@ -14,6 +14,7 @@
 
 #include "SampleUtils.h"
 
+#include "common/Assert.h"
 #include "common/Platform.h"
 #include "utils/BackendBinding.h"
 #include "wire/TerribleCommandBuffer.h"
@@ -212,8 +213,8 @@ bool InitSample(int argc, const char** argv) {
 
 void DoFlush() {
     if (cmdBufType == CmdBufType::Terrible) {
-        c2sBuf->Flush();
-        s2cBuf->Flush();
+        ASSERT(c2sBuf->Flush());
+        ASSERT(s2cBuf->Flush());
     }
     glfwPollEvents();
 }
--- a/src/tests/NXTTest.cpp
+++ b/src/tests/NXTTest.cpp
@@ -282,8 +282,8 @@ void NXTTest::SwapBuffersForCapture() {
 
 void NXTTest::FlushWire() {
     if (gTestUsesWire) {
-        mC2sBuf->Flush();
-        mS2cBuf->Flush();
+        ASSERT(mC2sBuf->Flush());
+        ASSERT(mS2cBuf->Flush());
     }
 }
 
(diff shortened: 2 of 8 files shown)


src/wire/TerribleCommandBuffer.cpp
```diff
@@ -43,7 +45,7 @@ namespace nxt { namespace wire {
     }
 
     void TerribleCommandBuffer::Flush() {
-        mHandler->HandleCommands(mBuffer, mOffset);
+        NXT_ASSERT(mHandler->HandleCommands(mBuffer, mOffset));
```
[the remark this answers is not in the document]
kainino0x: Do you think it should just return `bool` or should we use the Result.h/Error.h stuff out here?

kainino0x:
(sorry, current state of this is "not landable".) I changed some stuff to return bool, but it doesn't propagate enough to be useful. So I'll finish that up after your response on the above question:

> Do you think it should just return bool or should we use the Result.h/Error.h stuff out here?

src/wire/TerribleCommandBuffer.cpp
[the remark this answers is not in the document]
Kangz: I think we should just use bool instead of Result<void, void>. It is more self explanatory (though it doesn't explicitly show which of true or false is the success value).

Kangz:
No worries, and sorry for the slow review.

src/tests/unittests/WireTests.cpp
```diff
@@ -787,7 +788,7 @@ TEST_F(WireBufferMappingTests, UnmapCalledTooEarlyForRead) {
     nxtBufferUnmap(buffer);
 
     // The callback shouldn't get called, even when the request succeeded on the server side
-    FlushServer();
+    FlushServer(false);
```
kainino0x: Does this look like the right expectation? (note also that FlushClient on line 782 succeeds)

kainino0x:
(@Kangz PTAL at the last commit before I merge)

Kangz:
Figured out what the issue was. I pushed a fix for it to your branch and reverted the expectSuccess changes.

examples/SampleUtils.cpp
```diff
@@ -212,8 +213,8 @@ bool InitSample(int argc, const char** argv) {
 
 void DoFlush() {
     if (cmdBufType == CmdBufType::Terrible) {
-        c2sBuf->Flush();
-        s2cBuf->Flush();
+        NXT_ASSERT(c2sBuf->Flush());
```
Kangz: nit: or just ASSERT

src/tests/NXTTest.cpp
```diff
@@ -282,8 +282,8 @@ void NXTTest::SwapBuffersForCapture() {
 
 void NXTTest::FlushWire() {
     if (gTestUsesWire) {
-        mC2sBuf->Flush();
-        mS2cBuf->Flush();
+        NXT_ASSERT(mC2sBuf->Flush());
```
Kangz: ditto

kainino0x:
Thanks, your changes LGTM

PR #208, opened 2018-07-02
Manually retain/release mPendingCommands

kainino0x:
Fixes a crash for me on macOS+Metal

### DIFF
--- a/src/backend/metal/DeviceMTL.mm
+++ b/src/backend/metal/DeviceMTL.mm
@@ -66,7 +66,8 @@ namespace backend { namespace metal {
         }
         Tick();
 
-        mPendingCommands = nil;  // This will be autoreleased.
+        [mPendingCommands release];
+        mPendingCommands = nil;
 
         delete mMapTracker;
         mMapTracker = nullptr;
@@ -155,6 +156,7 @@ namespace backend { namespace metal {
     id<MTLCommandBuffer> Device::GetPendingCommandBuffer() {
         if (mPendingCommands == nil) {
             mPendingCommands = [mCommandQueue commandBuffer];
+            [mPendingCommands retain];
         }
         return mPendingCommands;
     }
@@ -174,7 +176,8 @@ namespace backend { namespace metal {
         }];
 
         [mPendingCommands commit];
-        mPendingCommands = nil;  // This will be autoreleased.
+        [mPendingCommands release];
+        mPendingCommands = nil;
         mPendingCommandSerial++;
     }
 


kainino0x:
@SenorBlanco I've recently found this is necessary to get demos and tests running on my laptop without crashing.

I was able to develop on this branch a week or so ago, so I actually don't know exactly what has changed. I have been changing my Xcode versions around a bit lately, so I'm suspecting that this affects certain Xcode versions in particular. I just rebuilt on Xcode 9.4 (9F1027a) and this patch is necessary there. It was also necessary with whatever version of Xcode is in `chrome/src/build/mac_files` (I think I was using 8.3.2 (8E2002)?)

(cc @Kangz)

src/backend/metal/DeviceMTL.mm
```diff
@@ -155,6 +156,7 @@ void Init(id<MTLDevice> metalDevice, nxtProcTable* procs, nxtDevice* device) {
     id<MTLCommandBuffer> Device::GetPendingCommandBuffer() {
         if (mPendingCommands == nil) {
             mPendingCommands = [mCommandQueue commandBuffer];
```
Kangz: The Metal headers says this command buffer is auto-released:

```
/*! 
 @method commandBuffer
 @abstract Returns a new autoreleased command buffer used to encode work into this queue that 
 maintains strong references to resources used within the command buffer.
*/
- (id <MTLCommandBuffer>)commandBuffer;
```

After more digging I finally understood what that means: the object is added to the current autorelease pool and `release` is called on it when the `NSAutoreleasePool` is drained, in our case every frame in [glfw](https://github.com/glfw/glfw/blob/master/src/cocoa_window.m#L1581). I think the samples that crash happen to create an `MTLCommandBuffer` before a glfwPollEvents and then tries to keep using afterwards, leading to a use after free.

Kangz:
I think I just didn't test the revert enough on the samples, sorry about that. On the plus side I found why this causes crashes :D

src/backend/metal/DeviceMTL.mm
SenorBlanco: Yes, that's my understanding too. So do we really want to manually release/retain mPendingCommands here? Or am I misunderstanding this patch?

Kangz: Yes we want to because the pending commands can be kept across "frame boundary" which is where the autoreleasing happens in our samples. You can see it in this example frameloop:

```cpp
while(true) {
  Create nxt::CommandBuffer
  nxt::Queue::Submit // mPendingCommands is submitted and set to nil
  nxt::Buffer::SetSubData // mPendingCommands is created to hold the copy from staging to GPU
  glfwPollEvent(); // autorelease, mPendingCommands is now invalid
}
```

SenorBlanco: Ah, that makes sense. Thanks!

SenorBlanco:
(Never mind, it wasn't after merge, I just got confused.)

PR #209, opened 2018-07-03
Allow bind group layouts to be sparse in pipeline layout

kainino0x:
Instead of initializing all of the bind group layouts in the pipeline
layout to default values, let them be nullptr and allow this elsewhere.

This is a follow-up on some changes in #206.

Also, rename GetBindGroupsLayoutMask -> GetBindGroupLayoutsMask

### DIFF
--- a/src/backend/PipelineLayout.cpp
+++ b/src/backend/PipelineLayout.cpp
@@ -49,7 +49,7 @@ namespace backend {
         return mBindGroupLayouts[group].Get();
     }
 
-    const std::bitset<kMaxBindGroups> PipelineLayoutBase::GetBindGroupsLayoutMask() const {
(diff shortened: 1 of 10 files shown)


src/backend/PipelineLayout.cpp
```diff
@@ -34,31 +34,22 @@ namespace backend {
 
     // PipelineLayoutBase
 
-    PipelineLayoutBase::PipelineLayoutBase(DeviceBase* device,
+    PipelineLayoutBase::PipelineLayoutBase(DeviceBase*,
```
Kangz: nit: Can we have the mDevice / GetDevice combo in PipelineLayoutBase too ?

kainino0x: I was thinking the same thing actually, I'll do that.

PR #210, opened 2018-07-03
Remove test samples

Kangz:
PTAL! Hopefully this removes some of the burden of converting to descriptors.


Kangz:
Thanks for the reviews!

PR #211, opened 2018-07-04
Use a descriptor for BindGroupLayout

kainino0x:


src/tests/unittests/WireTests.cpp
```diff
@@ -436,6 +441,30 @@ TEST_F(WireTests, StructureOfObjectArrayArgument) {
     FlushClient();
 }
 
+// Test that the wire is able to send structures that contain objects
+TEST_F(WireTests, StructureOfStructureArrayArgument) {
+    static constexpr int NUM_BINDINGS = 3;
+    nxtBindGroupBinding bindings[NUM_BINDINGS] {
+      { 0, NXT_SHADER_STAGE_BIT_VERTEX, NXT_BINDING_TYPE_SAMPLER, 2 },
+      { 1, NXT_SHADER_STAGE_BIT_VERTEX, NXT_BINDING_TYPE_SAMPLED_TEXTURE, 1 },
+      { 2, static_cast<nxtShaderStageBit>(NXT_SHADER_STAGE_BIT_VERTEX | NXT_SHADER_STAGE_BIT_FRAGMENT), NXT_BINDING_TYPE_UNIFORM_BUFFER, 1 },
```
kainino0x: the static_cast here is unfortunate, but not sure how to avoid it. (causes a type error :int where there should be an nxtShaderStageBit).

src/tests/unittests/WireTests.cpp
```diff
@@ -410,14 +410,17 @@ TEST_F(WireTests, StructureOfValuesArgument) {
...
-    EXPECT_CALL(api, DeviceCreateBindGroupLayoutBuilder(apiDevice))
-          .WillOnce(Return(apiBglBuilder));
+    nxtBindGroupLayout bgl = nxtDeviceCreateBindGroupLayout(device, &bglDescriptor);
     nxtBindGroupLayout apiBgl = api.GetNewBindGroupLayout();
-    EXPECT_CALL(api, BindGroupLayoutBuilderGetResult(apiBglBuilder))
+    EXPECT_CALL(api, DeviceCreateBindGroupLayout(apiDevice, MatchesLambda([](const nxtBindGroupLayoutDescriptor* desc) -> bool {
+        return desc->nextInChain == nullptr &&
+            desc->numBindings == 0 &&
+            desc->bindings == nullptr;
+    })))
```
Kangz: nit: This test is for the pipeline layout so we could just use the gmock wildcard `_`

src/backend/Device.cpp
```diff
@@ -98,8 +98,23 @@ namespace backend {
     BindGroupBuilder* DeviceBase::CreateBindGroupBuilder() {
         return new BindGroupBuilder(this);
     }
-    BindGroupLayoutBuilder* DeviceBase::CreateBindGroupLayoutBuilder() {
-        return new BindGroupLayoutBuilder(this);
+    BindGroupLayoutBase* DeviceBase::CreateBindGroupLayout(
```
Kangz: BindGroups are a bit different from other objects because they need to be deduplicated to allow comparison by pointer (helps make the check that BGs match the PL fast). The validation code part of the code would stay the same, then a BindGroupLayoutBase would be created with blueprint=true and finally GetOrCreateBindGroupLayout called.

Kangz: In the future we should add a validation test that a if BGL1 == BGL2 then a BG of layout BGL1 can be used with a PL slot with layout BGL2.

src/utils/NXTHelpers.cpp
```diff
@@ -162,4 +163,12 @@ namespace utils {
         return device.CreatePipelineLayout(&descriptor);
     }
 
+    nxt::BindGroupLayout MakeBindGroupLayout(
+        const nxt::Device& device, std::initializer_list<nxt::BindGroupBinding> bindings) {
```
Kangz: Nice!

examples/ComputeBoids.cpp
```diff
@@ -223,10 +223,11 @@ void initSim() {
...
     )");
 
-    nxt::BindGroupLayout bgl = device.CreateBindGroupLayoutBuilder()
-        .SetBindingsType(nxt::ShaderStageBit::Compute, nxt::BindingType::UniformBuffer, 0, 1)
-        .SetBindingsType(nxt::ShaderStageBit::Compute, nxt::BindingType::StorageBuffer, 1, 2)
-        .GetResult();
+    auto bgl = utils::MakeBindGroupLayout(device, {
+        { 0, nxt::ShaderStageBit::Compute, nxt::BindingType::UniformBuffer, 1 },
+        { 1, nxt::ShaderStageBit::Compute, nxt::BindingType::StorageBuffer, 1 },
+        { 2, nxt::ShaderStageBit::Compute, nxt::BindingType::StorageBuffer, 1 },
```
SenorBlanco: Just for my own edification: why not use a single entry with an array size of two here, as it was before, instead of two entries? Is the array size semantically different from the old "count"?

examples/glTFViewer/glTFViewer.cpp
```diff
@@ -271,15 +271,18 @@ namespace {
...
-                .SetBindingsType(nxt::ShaderStageBit::Fragment, nxt::BindingType::Sampler, 1, 1)
-                .SetBindingsType(nxt::ShaderStageBit::Fragment, nxt::BindingType::SampledTexture, 2, 1);
+            bindGroupLayout = utils::MakeBindGroupLayout(device, {
+                { 0, nxt::ShaderStageBit::Vertex, nxt::BindingType::UniformBuffer, 1 },
+                { 1, nxt::ShaderStageBit::Fragment, nxt::BindingType::Sampler, 1 },
+                { 2, nxt::ShaderStageBit::Fragment, nxt::BindingType::SampledTexture, 1 },
+            });
+        } else {
+            bindGroupLayout = utils::MakeBindGroupLayout(device, {
+                { 0, nxt::ShaderStageBit::Vertex, nxt::BindingType::UniformBuffer, 1 },
```
SenorBlanco: The code duplication here is a little unfortunate. Can't think of a clean way to fix it, though. (I'm no std::initializer_list expert, but I'm assuming there's no way to dynamically concatenate the optional ones.) Maybe a per-binding boolean which says whether to use that binding or not? Or use an array size of zero to signal an unused binding, and zero out the unused bindings' array sizes? Neither seems great, though.

src/tests/unittests/WireTests.cpp
```diff
@@ -418,8 +419,9 @@ TEST_F(WireTests, StructureOfObjectArrayArgument) {
     nxtBindGroupLayout apiBgl = api.GetNewBindGroupLayout();
     EXPECT_CALL(api, DeviceCreateBindGroupLayout(apiDevice, MatchesLambda([](const nxtBindGroupLayoutDescriptor* desc) -> bool {
         return desc->nextInChain == nullptr &&
-            desc->numBindings == 0 &&
-            desc->bindings == nullptr;
+            desc->numBindings == 0;
+            // desc->bindings isn't nullptr (its allocation is inline in the
```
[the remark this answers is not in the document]
kainino0x: It seems that way. I'll add a TODO

src/backend/Device.cpp
kainino0x: Oops, didn't realize I broke this. Fixed (and added test)

src/tests/unittests/WireTests.cpp
kainino0x: Done

src/tests/unittests/validation/BindGroupValidationTests.cpp
```diff
@@ -103,3 +105,17 @@ TEST_F(BindGroupValidationTest, BufferViewOffset) {
             .GetResult();
     }
 }
+
+TEST_F(BindGroupValidationTest, BindGroupLayoutCache) {
```
kainino0x: @Kangz PTAL at this new test in particular

src/tests/unittests/WireTests.cpp
```diff
@@ -436,6 +435,40 @@ TEST_F(WireTests, StructureOfObjectArrayArgument) {
...
+    bglDescriptor.bindings = bindings;
+
+    nxtDeviceCreateBindGroupLayout(device, &bglDescriptor);
+    nxtBindGroupLayout apiBgl = api.GetNewBindGroupLayout();
+    EXPECT_CALL(
+        api,
+        DeviceCreateBindGroupLayout(
+            apiDevice, MatchesLambda([bindings](const nxtBindGroupLayoutDescriptor* desc) -> bool {
+                for (int i = 0; i < NUM_BINDINGS; ++i) {
+                    if (memcmp(&desc->bindings[i], &bindings[i], sizeof(nxtBindGroupBinding)) !=
```
Kangz: Sorry this still could fail if the structure has gaps and they contain values that are different.

For example:

```
struct {
  char a;
  int b;
}
```

is actually

```
struct {
  char a;
  char gap[3];
  int b;
}
```

src/backend/BindGroupLayout.cpp
```diff
@@ -22,6 +22,21 @@
 
 namespace backend {
 
+    MaybeError ValidateBindGroupLayoutDescriptor(DeviceBase*,
+                                                 const nxt::BindGroupLayoutDescriptor* descriptor) {
+        NXT_TRY_ASSERT(descriptor->nextInChain == nullptr, "nextInChain must be nullptr");
```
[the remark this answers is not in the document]
Kangz: Nah it's fine, we'll have a path at making error messages more useful closer to a WebGPU v1.

src/tests/unittests/validation/BindGroupValidationTests.cpp
Kangz: This won't work if/when we run unittests through the wire. LGTM for now but can you add a comment explaining this only works without wire because we get the actual pointers?

examples/glTFViewer/glTFViewer.cpp
```diff
@@ -271,15 +271,13 @@ namespace {
...
-                .SetBindingsType(nxt::ShaderStageBit::Fragment, nxt::BindingType::Sampler, 1, 1)
-                .SetBindingsType(nxt::ShaderStageBit::Fragment, nxt::BindingType::SampledTexture, 2, 1);
-        }
-        auto bindGroupLayout = bindGroupLayoutBuilder.GetResult();
+        nxt::BindGroupLayout bindGroupLayout = utils::MakeBindGroupLayout(
+            device, {
+                        {0, nxt::ShaderStageBit::Vertex, nxt::BindingType::UniformBuffer, 1},
+                        {1, nxt::ShaderStageBit::Fragment, nxt::BindingType::Sampler, 1},
+                        {2, nxt::ShaderStageBit::Fragment, nxt::BindingType::SampledTexture,
+                         static_cast<uint32_t>(hasTexture ? 1 : 0)},
```
Kangz: This will fail validation currently

src/tests/unittests/WireTests.cpp
kainino0x: Ah, right, done

src/tests/unittests/validation/BindGroupValidationTests.cpp
kainino0x: Okay, done

examples/glTFViewer/glTFViewer.cpp
kainino0x: MakeBindGroupLayout - the helper function - special-cases the '0' arraySize, and doesn't record it into the BGL descriptor.

kainino0x:
I decided to remove the `arraySize` parameter for bind group layouts in this PR. I'm currently very unclear on what the behavior should be - in particular, I'm not sure whether this Vulkan semantic is even possible on all of the APIs. (I think it is, but it's not quite trivial to implement).

src/utils/NXTHelpers.cpp
```diff
@@ -162,4 +163,21 @@ namespace utils {
         return device.CreatePipelineLayout(&descriptor);
     }
 
+    nxt::BindGroupLayout MakeBindGroupLayout(
+        const nxt::Device& device,
+        std::initializer_list<nxt::BindGroupBinding> bindingsInitializer) {
+        std::vector<nxt::BindGroupBinding> bindings;
+        nxt::ShaderStageBit kNoStages{};
+        for (const nxt::BindGroupBinding& binding : bindingsInitializer) {
+            if (binding.visibility != kNoStages) {
```
kainino0x: `MakeBindGroupLayout` now uses a `visibility` of `0` to indicate bindings that should be skipped.

PR #212, opened 2018-07-04
Iterate per pass

Kangz:
PTAL. Sorry this moves a lot of code around but the changes to CommandBufferBuilder::ValidateGetResult and CommandBufferVk will help a bunch when implementing implicit barriers. Not having all passes in the same iteration does simplify some code overall too.

CommandBufferD3D12 is still TBD because I don't have a Windows machine. I should receive one soon though.


Kangz:
@kainino0x, @SenorBlanco PTAL! I'm starting to have a bunch of local commits depending on this.

src/backend/CommandBuffer.cpp
```diff
@@ -292,33 +385,25 @@ namespace backend {
                     // Validation of count and offset has already been done when the command was
                     // recorded because it impacts the size of an allocation in the
                     // CommandAllocator.
-                    if (!mState->ValidateSetPushConstants(cmd->stages)) {
+                    if (cmd->stages &
+                        ~(nxt::ShaderStageBit::Vertex | nxt::ShaderStageBit::Fragment)) {
+                        HandleError(
+                            "SetPushConstants stage must be a subset if (vertex|fragment) in "
```
kainino0x: "if" -> "of"

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -446,85 +374,249 @@ namespace backend { namespace opengl {
                     glDeleteFramebuffers(1, &readFBO);
                 } break;
 
+                case Command::TransitionBufferUsage: {
+                    TransitionBufferUsageCmd* cmd =
```
SenorBlanco: Not new, but "auto" here would make this less rendundant, and make it wrap better.

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -446,85 +374,249 @@ namespace backend { namespace opengl {
...
+                        mCommands.NextCommand<TransitionBufferUsageCmd>();
+
+                    cmd->buffer->UpdateUsageInternal(cmd->usage);
+                } break;
+
+                case Command::TransitionTextureUsage: {
+                    TransitionTextureUsageCmd* cmd =
```
SenorBlanco: Same here (and many places below).

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -446,85 +374,249 @@ namespace backend { namespace opengl {
...
+            unsigned int attachmentCount = 0;
+            for (uint32_t i : IterateBitSet(renderPass->GetColorAttachmentMask())) {
+                TextureViewBase* textureView = renderPass->GetColorAttachment(i).view.Get();
+                GLuint texture = ToBackend(textureView->GetTexture())->GetHandle();
+
+                // Attach color buffers.
+                glFramebufferTexture2D(GL_DRAW_FRAMEBUFFER, GL_COLOR_ATTACHMENT0 + i, GL_TEXTURE_2D,
+                                       texture, 0);
+                drawBuffers[i] = GL_COLOR_ATTACHMENT0 + i;
+                attachmentCount = i + 1;
```
SenorBlanco: Nit: not new, but doing this every iteration seems redundant. Not sure if this works with IterateBitSet(), but scope i outside the loop and use it as attachmentCount after?

src/backend/opengl/CommandBufferGL.cpp
```diff
@@ -446,85 +374,249 @@ namespace backend { namespace opengl {
...
+                case Command::EndRenderPass: {
+                    mCommands.NextCommand<EndRenderPassCmd>();
+                    glDeleteFramebuffers(1, &fbo);
+                    return;
+                } break;
+
                 case Command::DrawArrays: {
                     DrawArraysCmd* draw = mCommands.NextCommand<DrawArraysCmd>();
-                    pushConstants.Apply(lastPipeline, lastGLPipeline);
+                    pushConstants.Apply(lastPipeline, lastPipeline);
```
SenorBlanco: Apply() always seems to be called with lastPipeline twice now (although I could be wrong; didn't check non-diffed code). Consider removing one of the params if possible?

src/backend/CommandBuffer.cpp
SenorBlanco: Nit: "subset if" or "subset of"?

src/backend/metal/CommandBufferMTL.mm
```diff
@@ -29,111 +29,182 @@
...
                     [blit endEncoding];
                     blit = nil;  // This will be autoreleased.
                 }
             }
 
             void EnsureBlit(id<MTLCommandBuffer> commandBuffer) {
-                ASSERT(render == nil);
-                ASSERT(compute == nil);
                 if (blit == nil) {
+                    Finish();
```
kainino0x: Isn't this a no-op because (blit == nil) and (blit != nil)?

src/backend/opengl/CommandBufferGL.cpp
Kangz: Left as-is because we didn't decide yet wether render pass attachments have to be dense in [0, N] or can be sparse: I thought of using `renderPass->GetColorAttachmentMask().count()` but this might not be correct if render pass attachments are sparse.

src/backend/opengl/CommandBufferGL.cpp
Kangz: ditto

src/backend/CommandBuffer.cpp
Kangz: Done

Kangz: Done

src/backend/metal/CommandBufferMTL.mm
Kangz: Done. I'll add it back when we have multiple command encoders at the top-level (if we need them sometime)

Kangz:
Fixed commit message too.

Kangz:
Thanks for the review!

src/backend/opengl/CommandBufferGL.cpp
SenorBlanco: Never mind.. I see now this is necessary to force two different upcasts. I hate MI...

Kangz: Whoops sorry I forgot to answer to this. Yeah that's because of multiple inheritance :/.

PR #214, opened 2018-07-09
Fix ComputeBoids

kainino0x:
I know there was some churn on which version of this shader was correct - but right now, this goes from broken on all 4 backends to working on all 4 backends. So this seems better.

The old code (accidentally?) uses descriptor arrays (I think), which are not trivial to support on all backends, so we won't use them for now.

### DIFF
--- a/examples/ComputeBoids.cpp
+++ b/examples/ComputeBoids.cpp
@@ -154,12 +154,12 @@ void initSim() {
         } params;
 
         layout(std140, set = 0, binding = 1) buffer ParticlesA {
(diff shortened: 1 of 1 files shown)


kainino0x:
Right. Based on playing around with the tests in #216, arrays of descriptors are allowed in Vulkan (it seems to be what `descriptorCount` refers to), but not allowed in GL. (I got an error somewhere saying that specifically.)

As I mentioned in #211, we would need to investigate/discuss whether descriptor arrays should be in WebGPU.

PR #215, opened 2018-07-10
Roll glslang to 1ea8f595

kainino0x:
Rolling this to fix some memory leaks.


Kangz:
I'm surprised we don't need to roll spirv-tools at the same time, but since tests pass I guess it's fine.

kainino0x:
I was surprised as well. But everything seems good.

PR #217, opened 2018-07-10
Implement naive implicit barriers in the Vulkan backend.

Kangz:
PTAL, this makes the Vulkan backend ignore frontend hints about transition and perform all transitions on its own. The current implementation is quite naive, but has the advantage of being simple and correct (at least it passes all the tests).

FYI the next steps are to do the same for D3D12, then remove the concept of usage transitions from NXT, then cleanups/optimizations.

### DIFF
--- a/src/backend/CommandBuffer.cpp
+++ b/src/backend/CommandBuffer.cpp
@@ -122,6 +122,142 @@ namespace backend {
             return true;
         }
 
+        enum class PassType {
+            Render,
+            Compute,
+        };
+
+        // Helper class to encapsulate the logic of tracking per-resource usage during the
+        // validation of command buffer passes. It is used both to know if there are validation
+        // errors, and to get a list of resources used per pass for backends that need the
+        // information.
+        class PassResourceUsageTracker {
+          public:
+            void BufferUsedAs(BufferBase* buffer, nxt::BufferUsageBit usage) {
+                // std::map's operator[] will create the key and return 0 if the key didn't exist
+                // before.
+                nxt::BufferUsageBit& storedUsage = mBufferUsages[buffer];
+
+                if (usage == nxt::BufferUsageBit::Storage &&
+                    storedUsage & nxt::BufferUsageBit::Storage) {
+                    mStorageUsedMultipleTimes = true;
+                }
+
+                storedUsage |= usage;
+            }
+
+            void TextureUsedAs(TextureBase* texture, nxt::TextureUsageBit usage) {
+                // std::map's operator[] will create the key and return 0 if the key didn't exist
+                // before.
+                nxt::TextureUsageBit& storedUsage = mTextureUsages[texture];
+
+                if (usage == nxt::TextureUsageBit::Storage &&
+                    storedUsage & nxt::TextureUsageBit::Storage) {
+                    mStorageUsedMultipleTimes = true;
+                }
+
+                storedUsage |= usage;
+            }
+
+            // Performs the per-pass usage validation checks
+            bool AreUsagesValid(PassType pass) const {
+                // Storage resources cannot be used twice in the same compute pass
+                if (pass == PassType::Compute && mStorageUsedMultipleTimes) {
+                    return false;
+                }
+
+                // Buffers can only be used as single-write or multiple read.
+                for (auto& it : mBufferUsages) {
+                    BufferBase* buffer = it.first;
+                    nxt::BufferUsageBit usage = it.second;
+
+                    if (usage & ~buffer->GetAllowedUsage()) {
+                        return false;
+                    }
+
+                    bool readOnly = (usage & kReadOnlyBufferUsages) == usage;
+                    bool singleUse = nxt::HasZeroOrOneBits(usage);
+
+                    if (!readOnly && !singleUse) {
+                        return false;
+                    }
+                }
+
+                // Textures can only be used as single-write or multiple read.
+                // TODO(cwallez@chromium.org): implement per-subresource tracking
+                for (auto& it : mTextureUsages) {
+                    TextureBase* texture = it.first;
+                    nxt::TextureUsageBit usage = it.second;
+
+                    if (usage & ~texture->GetAllowedUsage()) {
+                        return false;
+                    }
+
+                    // For textures the only read-only usage in a pass is Sampled, so checking the
+                    // usage constraint simplifies to checking a single usage bit is set.
+                    if (!nxt::HasZeroOrOneBits(it.second)) {
+                        return false;
+                    }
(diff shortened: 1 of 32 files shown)


src/backend/CommandBuffer.cpp
```diff
@@ -122,6 +122,137 @@ namespace backend {
...
+                if (usage == nxt::TextureUsageBit::Storage &&
+                    storedUsage & nxt::TextureUsageBit::Storage) {
+                    mStorageUsedMultipleTimes = true;
+                }
+
+                storedUsage |= usage;
+            }
+
+            // Performs the per-pass usage validation checks
+            bool AreUsagesValid(bool isCompute) const {
```
SenorBlanco: Prefer an enum here (maybe ShaderStageBit, although that's not exactly right) instead, so the function doesn't take an anonymous bool.

src/backend/CommandBuffer.cpp
```diff
@@ -171,6 +302,12 @@ namespace backend {
         return std::move(mIterator);
     }
 
+    std::vector<PassResourceUsage> CommandBufferBuilder::AcquirePassResourceUsage() {
+        ASSERT(!mWerePassUsagesAcquired);
+        mWerePassUsagesAcquired = true;
+        return std::move(mPassResourceUsages);
```
SenorBlanco: uNit: "return std::move()" used here but not above in AcquireResourceUsage() where you return by value. Better to be consistent and use it in both places or neither. (I *think* it's superfluous, but I'm still getting up to speed.)

src/backend/d3d12/DeviceD3D12.cpp
```diff
@@ -51,7 +51,12 @@ namespace backend { namespace d3d12 {
 
     nxtSwapChainImplementation CreateNativeSwapChainImpl(nxtDevice device, HWND window) {
         Device* backendDevice = reinterpret_cast<Device*>(device);
-        return CreateSwapChainImplementation(new NativeSwapChainImpl(backendDevice, window));
+
+        nxtSwapChainImplementation impl;
+        impl = CreateSwapChainImplementation(new NativeSwapChainImpl(backendDevice, window));
+        impl.textureUsage = NXT_TEXTURE_USAGE_BIT_PRESENT;
```
SenorBlanco: Would it be easier to simply make this bit the default for all backends, and ignore it for the backends which don't require barriers?

src/include/nxt/EnumClassBitmasks.h
```diff
@@ -53,16 +53,23 @@ namespace nxt {
     struct LowerBitmask<T, typename std::enable_if<IsNXTBitmask<T>::enable>::type> {
         static constexpr bool enable = true;
         using type = T;
-        static T Lower(T t) {return t;}
+        constexpr static T Lower(T t) {
```
SenorBlanco: If possible, could you land this commit separately, since it's just constexpr and whitespace changes? I see it's in the middle, but you could probably reorder and squash the other two.

src/backend/CommandBuffer.cpp
Kangz: Without the `std::move` here, a whole copy of the `PassResourceUsage` is created. `AcquireResourceUsage` doesn't need it because it takes advantage of guaranteed return value optimization.

src/backend/d3d12/DeviceD3D12.cpp
Kangz: We might have swapchains that don't present, for example Chromium's compositor that would rather have `NXT_TEXTURE_USAGE_BIT_SAMPLED`.

src/include/nxt/EnumClassBitmasks.h
Kangz: Done

src/backend/CommandBuffer.cpp
SenorBlanco: Out of curiosity, why is RVO guaranteed there, but not here?

src/backend/d3d12/DeviceD3D12.cpp
SenorBlanco: Ah, that makes sense.

src/backend/CommandBuffer.cpp
Kangz: There we have the return value allocated inside the function and a single return at the end. In the calling convention, non-POD types are returned via a hidden reference: `Type f(int a)` gets converted to `void f(Type* result, int a)`. Then the compiler is able to allocate the return value directly into the "return" pointer, skipping the copy altogether. Using std::move in that case is actually a deoptimization because it forces the move constructor to be called, when it isn't needed.

For the case with std::move, the variable already exists as a member of the CommandBufferBuilder so we must have either a copy construction or a move construction (or a construction and manual copying / moving). RVO is not possible since the variable already exist, hence the move.

src/backend/vulkan/TextureVk.cpp
```diff
@@ -312,22 +306,24 @@ namespace backend { namespace vulkan {
...
     }
 
-    // Helper function to add a texture barrier to a command buffer. This is inefficient because we
-    // should be coalescing barriers as much as possible.
-    void Texture::RecordBarrier(VkCommandBuffer commands,
-                                nxt::TextureUsageBit currentUsage,
-                                nxt::TextureUsageBit targetUsage) const {
+    void Texture::TransitionUsageNow(VkCommandBuffer commands, nxt::TextureUsageBit usage) {
+        // Avoid encoding barriers when it isn't needed.
+        if (mLastUsage == usage) {
```
kainino0x: Is a barrier not needed if transitioning from a write usage to a write usage?

src/backend/CommandBuffer.cpp
SenorBlanco: Ah of course, thanks!

src/backend/CommandBuffer.cpp
```diff
@@ -122,6 +122,142 @@ namespace backend {
...
+        enum class PassType {
+            Render,
+            Compute,
+        };
+
+        // Helper class to encapsulate the logic of tracking per-resource usage during the
+        // validation of command buffer passes. It is used both to know if there are validation
+        // errors, and to get a list of resources used per pass for backends that need the
+        // information.
+        class PassResourceUsageTracker {
```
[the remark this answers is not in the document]
Kangz: :D

src/backend/vulkan/TextureVk.cpp
Kangz: Done

Kangz:
Thanks for the reviews!

PR #218, opened 2018-07-10
Fix D3D12 descriptor renumbering

kainino0x:
Fixes ComputeCopyStorageBufferTests.BasicTest/D3D12


src/backend/d3d12/ShaderModuleD3D12.cpp
```diff
@@ -33,26 +33,24 @@ namespace backend { namespace d3d12 {
...
+        const auto& moduleBindingInfo = GetBindingInfo();
+        for (int group = 0; group < moduleBindingInfo.size(); ++group) {
+            const auto& groupBindingInfo = moduleBindingInfo[group];
+
+            static constexpr int kNumBindingTypes = 4;
+            std::array<uint32_t, kNumBindingTypes> baseRegisters{};
+
+            for (const auto& bindingInfo : groupBindingInfo) {
+                if (bindingInfo.used) {
+                    auto& baseRegister = baseRegisters[static_cast<uint32_t>(bindingInfo.type)];
```
Kangz: Can we have a switch instead of a static_cast, or at least have an assert that we are inside the array?

Also I think we should use auto a bit less: Chromium's style guide asks that simple types are spelled out.

kainino0x: I wish we had an `enum_map` type. For now I added an assert.

kainino0x: Actually changed it to a switch.

PR #219, opened 2018-07-11
Implement naive implicit barriers on the D3D12 backend.

Kangz:
PTAL!

### DIFF
--- a/src/backend/d3d12/BufferD3D12.cpp
+++ b/src/backend/d3d12/BufferD3D12.cpp
@@ -67,7 +67,7 @@ namespace backend { namespace d3d12 {
         }
     }  // namespace
 
-    Buffer::Buffer(Device* device, BufferBuilder* builder) : BufferBase(builder), mDevice(device) {
+    Buffer::Buffer(BufferBuilder* builder) : BufferBase(builder) {
         D3D12_RESOURCE_DESC resourceDescriptor;
         resourceDescriptor.Dimension = D3D12_RESOURCE_DIMENSION_BUFFER;
         resourceDescriptor.Alignment = 0;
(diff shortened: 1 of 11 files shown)


src/backend/d3d12/BufferD3D12.cpp
```diff
@@ -113,31 +118,35 @@ namespace backend { namespace d3d12 {
...
+            return;
         }
 
-        barrier->Type = D3D12_RESOURCE_BARRIER_TYPE_TRANSITION;
-        barrier->Flags = D3D12_RESOURCE_BARRIER_FLAG_NONE;
-        barrier->Transition.pResource = mResource.Get();
-        barrier->Transition.StateBefore = stateBefore;
-        barrier->Transition.StateAfter = stateAfter;
-        barrier->Transition.Subresource = 0;
+        D3D12_RESOURCE_STATES lastState = D3D12BufferUsage(mLastUsage);
```
SenorBlanco: Nit (or personal style, and not new to this patch): it'd be nice if this helper func was ToD3D12BufferUsage(), so it didn't look like a constructor.

src/backend/d3d12/BufferD3D12.cpp
```diff
@@ -153,14 +162,19 @@ namespace backend { namespace d3d12 {
...
+        TransitionUsageNow(device->GetPendingCommandList(), nxt::BufferUsageBit::TransferDst);
+        device->GetResourceUploader()->BufferSubData(mResource, start, count, data);
     }
 
     void Buffer::MapReadAsyncImpl(uint32_t serial, uint32_t start, uint32_t count) {
         D3D12_RANGE readRange = {start, start + count};
         char* data = nullptr;
         ASSERT_SUCCESS(mResource->Map(0, &readRange, reinterpret_cast<void**>(&data)));
 
+        // There is no need to transition the resource to a new state: D3D12 seems to make the GPU
```
SenorBlanco: Not an entirely naive implementation, then. :)

src/backend/d3d12/CommandBufferD3D12.cpp
```diff
@@ -226,130 +226,54 @@ namespace backend { namespace d3d12 {
...
+
+            ID3D12DescriptorHeap* descriptorHeaps[2] = {
+                bindingTracker.cbvSrvUavGPUDescriptorHeap.Get(),
+                bindingTracker.samplerGPUDescriptorHeap.Get()};
+            if (descriptorHeaps[0] && descriptorHeaps[1]) {
+                commandList->SetDescriptorHeaps(2, descriptorHeaps);
+            } else if (descriptorHeaps[0]) {
+                commandList->SetDescriptorHeaps(1, descriptorHeaps);
+            } else if (descriptorHeaps[1]) {
+                commandList->SetDescriptorHeaps(2, &descriptorHeaps[1]);
```
kainino0x: Not new, but is this valid? Isn't this saying "array of 2 descriptor heaps starting at index 1" (in a size-2 array)?

Kangz: Done

src/backend/d3d12/BufferD3D12.cpp
Kangz: Yeah but wait until you rerun the skia-nxt benchmarks on Vulkan ^^

src/backend/d3d12/BufferD3D12.cpp
Kangz: I wouldn't mind that change but it would be best to do it across all backends at once so I'll leave it for later.

PR #220, opened 2018-07-13
Partially fix ComputeCopyStorageBufferTests.DISABLED_UnsizedArray

kainino0x:
And update comments on several ComputeCopyStorageBufferTests cases.

### DIFF
--- a/src/tests/end2end/ComputeCopyStorageBufferTests.cpp
+++ b/src/tests/end2end/ComputeCopyStorageBufferTests.cpp
@@ -109,6 +109,8 @@ TEST_P(ComputeCopyStorageBufferTests, BasicTest) {
 }
 
 // Test that a slightly-less-trivial compute-shader memcpy implementation works.
+//
+// TODO(kainino@chromium.org): Fails on D3D12 backend. Probably due to a limitation in SPIRV-Cross?
 TEST_P(ComputeCopyStorageBufferTests, StructTest) {
     BasicTest(R"(
         #version 450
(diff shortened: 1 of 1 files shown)


src/tests/end2end/ComputeCopyStorageBufferTests.cpp
```diff
@@ -145,11 +148,14 @@ TEST_P(ComputeCopyStorageBufferTests, DISABLED_SizedArray) {
 }
 
 // Test with an unsized array SSBO.
+//
+// TODO(kainino@chromium.org): This test may be somewhat wrong. I'm not sure whether this is
+// supposed to be possible on the various native APIs.
+// Linking on OpenGL fails with "OpenGL requires constant indexes for unsized array access(dst)".
```
Kangz: I think the unsized array has to be the last member of a structure that's the root of the SSBO. 

```
layout(std140, set=0, binding=0) buffer Src {S s[];} src;
```

kainino0x: This test would be an unsized array of descriptors, not a single descriptor referring to an unsized buffer. Again, I'm not sure it's possible.

I should probably add a test for an unsized buffer though.

kainino0x: Updated the test descriptions to be more accurate (now that I understand what they should be doing).

kainino0x: Also updated test names and added the unsized buffer test.

src/tests/end2end/ComputeCopyStorageBufferTests.cpp
```diff
@@ -109,7 +109,9 @@ TEST_P(ComputeCopyStorageBufferTests, BasicTest) {
 }
 
 // Test that a slightly-less-trivial compute-shader memcpy implementation works.
-TEST_P(ComputeCopyStorageBufferTests, StructTest) {
+//
+// TODO(kainino@chromium.org): Fails on D3D12 backend. Probably due to a limitation in SPIRV-Cross?
+TEST_P(ComputeCopyStorageBufferTests, SizedArrayOfStruct) {
```
Kangz: If it fails on some backends, can we skip it like we skip some tests in BufferTests?

Kangz:
LGTM + please squash

PR #221, opened 2018-07-13
Remove explicit usage transition from the API and validation

Kangz:
This removes the following for both Buffer and Texture:

 - The builder's SetInitialUsage
 - The object's FreezeUsage and TransitionUsage methods
 - The CommandBuffer Transition<Object>Usage methods

All samples and tests are simplified as a result. This also obsoletes
the UsageValidationTest which is removed.

Some validation was dependent on "current usage" and hasn't been
reintroduced for implicit transitions yet:

 - Buffers can be used while mapped
 - Swapchain textures can be used after they have been presented.

Validation for these will involve collecting all the resources used by a
command buffer and will be done in a follow-up patch.


src/backend/Buffer.cpp
```diff
@@ -155,58 +148,31 @@ namespace backend {
...
-        bool singleUse = nxt::HasZeroOrOneBits(usage);
-        return allowed && (readOnly || singleUse);
-    }
-
-    bool BufferBase::IsTransitionPossible(nxt::BufferUsageBit usage) const {
-        if (mIsFrozen || mIsMapped) {
+    bool BufferBase::ValidateMapBase(uint32_t start,
+                                     uint32_t size,
+                                     nxt::BufferUsageBit requiredUsage) {
+        if (start + size > GetSize()) {
```
kainino0x: Old code, but add a TODO or DCHECK about overflow here?

src/backend/Buffer.cpp
```diff
@@ -155,58 +148,31 @@ namespace backend {
...
+
+        return true;
     }
 
     // BufferBuilder
 
     enum BufferSetProperties {
         BUFFER_PROPERTY_ALLOWED_USAGE = 0x1,
-        BUFFER_PROPERTY_INITIAL_USAGE = 0x2,
         BUFFER_PROPERTY_SIZE = 0x4,
```
kainino0x: Any reason to keep this as 0x4 instead of 0x2?

src/backend/d3d12/TextureD3D12.cpp
```diff
@@ -122,9 +122,8 @@ namespace backend { namespace d3d12 {
         resourceDescriptor.Layout = D3D12_TEXTURE_LAYOUT_UNKNOWN;
         resourceDescriptor.Flags = D3D12ResourceFlags(GetAllowedUsage(), GetFormat());
 
-        mResource =
-            mDevice->GetResourceAllocator()->Allocate(D3D12_HEAP_TYPE_DEFAULT, resourceDescriptor,
-                                                      D3D12TextureUsage(GetUsage(), GetFormat()));
+        mResource = mDevice->GetResourceAllocator()->Allocate(
```
SenorBlanco: Could you explain this change a bit?

src/backend/CommandBuffer.cpp
```diff
@@ -351,11 +333,17 @@ namespace backend {
...
                     if (!ValidateCopySizeFitsInBuffer(this, copy->source, copy->size) ||
-                        !ValidateCopySizeFitsInBuffer(this, copy->destination, copy->size) ||
-                        !mState->ValidateCanUseBufferAs(copy->source.buffer.Get(),
-                                                        nxt::BufferUsageBit::TransferSrc) ||
-                        !mState->ValidateCanUseBufferAs(copy->destination.buffer.Get(),
-                                                        nxt::BufferUsageBit::TransferDst)) {
+                        !ValidateCopySizeFitsInBuffer(this, copy->destination, copy->size)) {
+                        return false;
+                    }
+                    if (!(copy->source.buffer->GetAllowedUsage() &
```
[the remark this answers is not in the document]
Kangz: Added helpers ValidateCanUseAs(this, buffer, usage) and a similar overload for textures.

src/backend/Buffer.cpp
Kangz: Done

src/backend/d3d12/TextureD3D12.cpp
Kangz: The change is that `D3D12TextureUsage(GetUsage(), GetFormat())` becomes `D3D12_RESOURCE_STATE_COMMON`. This is used by the resource allocator to pass as the `InitialResourceState` argument of [`ID3D12Device::CreateCommittedResource`](https://docs.microsoft.com/en-us/windows/desktop/api/d3d12/nf-d3d12-id3d12device-createcommittedresource).

Previously this would set the resource state to the one corresponding to the initial usage.Since we don't have initial usages anymore, we will always generate a resource state transition (until we optimize that) so we put a state that's always valid.

src/backend/Buffer.cpp
Kangz: Done

Kangz:
@kainino0x I think I know how to give you access to AppVeyor (omg their permission model is weird). Can you PM me your Github email address so I can add you?

Kangz:
Thanks for the reviews, landing this :D

src/backend/d3d12/TextureD3D12.cpp
SenorBlanco: Makes sense, thanks!

PR #223, opened 2018-07-16
Use Error.h error handling in parts of the frontend.

Kangz:
PTAL!

This is a refactor that's part of the cleanup of CommandBufferBuilder::ValidateGetResult after the implicit barriers refactor. Following up will be a cleanup of CommandBufferStateTracker, then probably optimizations of the usage tracking.


src/backend/Device.cpp
```diff
@@ -93,25 +93,20 @@ namespace backend {
...
     }
     BindGroupLayoutBase* DeviceBase::CreateBindGroupLayout(
         const nxt::BindGroupLayoutDescriptor* descriptor) {
-        MaybeError validation = ValidateBindGroupLayoutDescriptor(this, descriptor);
-        if (validation.IsError()) {
-            // TODO(cwallez@chromium.org): Implement the WebGPU error handling mechanism.
-            delete validation.AcquireError();
-            return nullptr;
-        }
+        BindGroupLayoutBase* result = nullptr;
```
SenorBlanco: uNit: spurious initializer? If it's never initialized by CreateBindGroupLayoutInternal(), it looks like it's never used, so should be safe to skip.

src/backend/Device.cpp
```diff
@@ -133,30 +128,22 @@ namespace backend {
...
-        return maybePipelineLayout.AcquireSuccess();
+
+        return result;
     }
     QueueBase* DeviceBase::CreateQueue() {
-        ResultOrError<QueueBase*> maybeQueue = CreateQueueImpl();
-        if (maybeQueue.IsError()) {
-            // TODO(cwallez@chromium.org): Implement the WebGPU error handling mechanism.
-            delete maybeQueue.AcquireError();
+        QueueBase* result = nullptr;
```
SenorBlanco: Ibid.

src/backend/Device.cpp
kainino0x: Do you mean "if it's always initialized"?

I personally prefer to be a bit defensive about cases like this, in case we somehow end up reaching `return result` without initializing `result`. Plus I like to hope that the compiler can elide this initializer.

Kangz: I don't think the compiler can elide it if it doesn't inline the `Internal` call. It should be super super negligeable compared to the rest of object creation.

Kangz:
PTAL at the Windows.h related commit that I just added.

Kangz:
Thanks!

PR #224, opened 2018-07-16
Fix glTFViewer by using push constants

kainino0x:
* Works on Metal
* Not working on OpenGL (black screen; not sure why)
* Not working on Vulkan or D3D12 (assertion crash; push constants aren't implemented?)

### DIFF
--- a/examples/glTFViewer/glTFViewer.cpp
+++ b/examples/glTFViewer/glTFViewer.cpp
@@ -64,7 +64,6 @@ namespace gl {
 }
 
 struct MaterialInfo {
-    nxt::Buffer uniformBuffer;
     nxt::RenderPipeline pipeline;
     nxt::BindGroup bindGroup0;
     std::map<uint32_t, std::string> slotSemantics;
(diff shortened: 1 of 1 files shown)


examples/glTFViewer/glTFViewer.cpp
```diff
@@ -273,10 +272,9 @@ namespace {
         constexpr nxt::ShaderStageBit kNoStages{};
         nxt::BindGroupLayout bindGroupLayout = utils::MakeBindGroupLayout(
             device, {
-                        {0, nxt::ShaderStageBit::Vertex, nxt::BindingType::UniformBuffer},
-                        {1, hasTexture ? nxt::ShaderStageBit::Fragment : kNoStages,
+                        {0, hasTexture ? nxt::ShaderStageBit::Fragment : kNoStages,
                          nxt::BindingType::Sampler},
-                        {2, hasTexture ? nxt::ShaderStageBit::Fragment : kNoStages,
+                        {1, hasTexture ? nxt::ShaderStageBit::Fragment : kNoStages,
```
kainino0x: It's actually possible to get rid of the entire bind group, but it's kind of annoying without std::optional, so I didn't bother.

Kangz: Ack, np

PR #226, opened 2018-07-18
Rename

Kangz:
PTAL this removes all the references to NXT except in the Readme.md files. I can squash into a couple commits if you think it is better.


Kangz:
Landing this now before the WebGPU meeting.

PR #227, opened 2018-07-18
Fix bind group binding ID validation.

SenorBlanco:
Use the binding, not the iterator index.


Kangz:
LGTM, there's a formatting error.

PR #228, opened 2018-07-19
State tracker

Kangz:
PTAL

### DIFF
--- a/src/backend/CommandBufferStateTracker.cpp
+++ b/src/backend/CommandBufferStateTracker.cpp
@@ -15,69 +15,124 @@
 #include "backend/CommandBufferStateTracker.h"
 
 #include "backend/BindGroup.h"
-#include "backend/BindGroupLayout.h"
-#include "backend/Buffer.h"
 #include "backend/ComputePipeline.h"
 #include "backend/Forward.h"
 #include "backend/InputState.h"
 #include "backend/PipelineLayout.h"
-#include "backend/RenderPassDescriptor.h"
 #include "backend/RenderPipeline.h"
-#include "backend/Texture.h"
 #include "common/Assert.h"
 #include "common/BitSetIterator.h"
 
 namespace backend {
 
-    MaybeError CommandBufferStateTracker::ValidateCanDispatch() {
-        constexpr ValidationAspects requiredAspects =
-            1 << VALIDATION_ASPECT_PIPELINE | 1 << VALIDATION_ASPECT_BIND_GROUPS;
-        if ((requiredAspects & ~mAspects).none()) {
-            // Fast return-true path if everything is good
-            return {};
-        }
+    enum ValidationAspect {
+        VALIDATION_ASPECT_PIPELINE,
+        VALIDATION_ASPECT_BIND_GROUPS,
+        VALIDATION_ASPECT_VERTEX_BUFFERS,
+        VALIDATION_ASPECT_INDEX_BUFFER,
 
-        if (!mAspects[VALIDATION_ASPECT_PIPELINE]) {
-            DAWN_RETURN_ERROR("No active compute pipeline");
-        }
-        // Compute the lazily computed mAspects
-        if (!RecomputeHaveAspectBindGroups()) {
-            DAWN_RETURN_ERROR("Bind group state not valid");
-        }
-        return {};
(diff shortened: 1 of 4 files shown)


src/backend/CommandBufferStateTracker.cpp
```diff
@@ -15,69 +15,123 @@
...
-            1 << VALIDATION_ASPECT_PIPELINE | 1 << VALIDATION_ASPECT_BIND_GROUPS |
-            1 << VALIDATION_ASPECT_VERTEX_BUFFERS | 1 << VALIDATION_ASPECT_INDEX_BUFFER;
-        if ((requiredAspects & ~mAspects).none()) {
-            // Fast return-true path if everything is good
-            return {};
+    MaybeError CommandBufferStateTracker::GenerateAspectError(ValidationAspects aspects) {
+        ASSERT(aspects.any());
+
+        if (aspects[VALIDATION_ASPECT_INDEX_BUFFER]) {
+            DAWN_RETURN_ERROR("Missing index buffer");
```
SenorBlanco: Is it the case that only the first failing aspect will be displayed?

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -15,69 +15,123 @@
...
+    void CommandBufferStateTracker::RecomputeLazyAspects(ValidationAspects aspects) {
+        ASSERT(mAspects[VALIDATION_ASPECT_PIPELINE]);
+        ASSERT((aspects & ~kLazyAspects).none());
+
+        if (aspects[VALIDATION_ASPECT_BIND_GROUPS]) {
+            bool matches = true;
+
+            for (uint32_t i : IterateBitSet(mLastPipelineLayout->GetBindGroupLayoutsMask())) {
+                if (mLastPipelineLayout->GetBindGroupLayout(i) != mBindgroups[i]->GetLayout()) {
+                    matches = false;
```
SenorBlanco: Nit: correct/worthwhile to break; here?

src/backend/CommandBufferStateTracker.cpp
Kangz: Yes, we'll want to have a making great error messages but that would be between the MVP and v1 I think.

src/backend/CommandBufferStateTracker.cpp
Kangz: Done

src/backend/CommandBufferStateTracker.cpp
SenorBlanco: Acknowledged.

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -90,96 +145,25 @@ namespace backend {
...
     }
 
     void CommandBufferStateTracker::SetPipelineCommon(PipelineBase* pipeline) {
-        PipelineLayoutBase* layout = pipeline->GetLayout();
+        mLastPipelineLayout = pipeline->GetLayout();
 
         mAspects.set(VALIDATION_ASPECT_PIPELINE);
-
         mAspects.reset(VALIDATION_ASPECT_BIND_GROUPS);
         mAspects.reset(VALIDATION_ASPECT_VERTEX_BUFFERS);
```
kainino0x: Probably an existing issue, but should this also reset `VALIDATION_ASPECT_INDEX_BUFFER`?

src/backend/CommandBufferStateTracker.cpp
```diff
@@ -15,69 +15,124 @@
...
+    MaybeError CommandBufferStateTracker::ValidateOperation(ValidationAspects requiredAspects) {
+        // Fast return-true path if everything is good
+        ValidationAspects missingAspects = requiredAspects & ~mAspects;
+        if (missingAspects.none()) {
             return {};
         }
 
-        if (!mAspects[VALIDATION_ASPECT_PIPELINE]) {
-            DAWN_RETURN_ERROR("No active compute pipeline");
+        // Generate an error immediatly if a non-lazy aspect is missing as computing lazy aspects
```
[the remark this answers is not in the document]
Kangz: Done

Kangz:
Yes it does loosen some validation that seemed it shouldn't have been there in the first place.

PR #229, opened 2018-07-19
BufferBase: Split validation in helper methods

Kangz:
PTAL

### DIFF
--- a/src/backend/Buffer.cpp
+++ b/src/backend/Buffer.cpp
@@ -78,13 +78,7 @@ namespace backend {
     }
 
     void BufferBase::SetSubData(uint32_t start, uint32_t count, const uint8_t* data) {
-        if (start + count > GetSize()) {
-            mDevice->HandleError("Buffer subdata out of range");
-            return;
-        }
-
-        if (!(mAllowedUsage & dawn::BufferUsageBit::TransferDst)) {
-            mDevice->HandleError("Buffer needs the transfer dst usage bit");
+        if (mDevice->ConsumedError(ValidateSetSubData(start, count))) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/backend/Buffer.cpp
```diff
@@ -148,26 +141,43 @@ namespace backend {
         mMapUserdata = 0;
     }
 
-    bool BufferBase::ValidateMapBase(uint32_t start,
-                                     uint32_t size,
-                                     dawn::BufferUsageBit requiredUsage) {
+    MaybeError BufferBase::ValidateSetSubData(uint32_t start, uint32_t count) const {
+        if (start + count > GetSize()) {
+            DAWN_RETURN_ERROR("Buffer needs the transfer dst usage bit");
```
SenorBlanco: This error message looks wrong.

Kangz: Done

src/backend/Buffer.cpp
```diff
@@ -148,26 +141,43 @@ namespace backend {
         mMapUserdata = 0;
     }
 
-    bool BufferBase::ValidateMapBase(uint32_t start,
-                                     uint32_t size,
-                                     dawn::BufferUsageBit requiredUsage) {
+    MaybeError BufferBase::ValidateSetSubData(uint32_t start, uint32_t count) const {
+        if (start + count > GetSize()) {
```
kainino0x: Handle int overflow? (or add a TODO)

src/backend/Buffer.cpp
```diff
@@ -148,26 +141,43 @@ namespace backend {
...
+        }
+
+        return {};
+    }
+
+    MaybeError BufferBase::ValidateMap(uint32_t start,
+                                       uint32_t size,
+                                       dawn::BufferUsageBit requiredUsage) const {
         // TODO(cwallez@chromium.org): check for overflows.
         if (start + size > GetSize()) {
```
kainino0x: Here too.

src/backend/Buffer.cpp
Kangz: Added a TODO

src/backend/Buffer.cpp
Kangz: There was a TODO already.

PR #232, opened 2018-07-26
Make libdawn_native a shared library.

Kangz:
PTAL. This is a bit more complicated than I expected because it forces us to sort-out our dependencies. (previously they leaked between binaries and dawn_native, making it compile when deps weren't set up properly).


src/dawn_native/dawn_platform.h
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+#ifndef DAWNNATIVE_DAWNPLATFORM_H_
+#define DAWNNATIVE_DAWNPLATFORM_H_
+
+// Use dawncpp to have the enum and bitfield definitions
+#include <dawn/dawncpp.h>
+
+// Use our autogenerated version of the dawn structures that point to dawn_native object types
+// (nxt::Buffer is dawn_native::BufferBase*)
```
SenorBlanco: s/nxt::Buffer/dawn::Buffer/ ?

third_party/CMakeLists.txt
```diff
@@ -138,5 +140,8 @@ set(STB_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/stb PARENT_SCOPE)
 # glm matrix math library
 set(GLM_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/glm PARENT_SCOPE)
 
+# glm matrix math library
```
kainino0x: wrong comment here

src/common/vulkan_platform.h
```diff
@@ -59,26 +61,62 @@ class VkNonDispatchableHandle {
         return mHandle != 0;
     }
 
-    static VkNonDispatchableHandle<Tag> CreateFromHandle(uint64_t handle) {
+    // The regular Vulkan handle type depends on the pointer width but is always 64 bits wide.
```
kainino0x: This reads as "The regular Vulkan handle type depends on the pointer width" + "The regular Vulkan handle type is always 64 bits wide" - sounds like a contradiction.

kainino0x:
Also needs reformat

src/dawn_native/dawn_platform.h
Kangz: Done

third_party/CMakeLists.txt
Kangz: Done

src/common/vulkan_platform.h
Kangz: Done, explained better what this means.

Kangz:
We don't have a good description of what the different targets are, I'll rework the README after the shared lib work.

PR #233, opened 2018-07-27
Make dawn_wire a shared library.

Kangz:
PTAL this is the last library that needed to be a shared library!


Kangz:
Finally figured out the link issue with glad. Link included in the comment in third_party/CMakeLists.txt and opened an issue against glad.

PTAL again since I also added a commit that makes the shared library dependencies private.

Kangz:
Thanks for the reviews!

PR #235, opened 2018-08-01
Preliminary changes for the GN build

Kangz:
These are a bunch of changes that help make BUILD.gn files for Dawn. PTAL!

Most notably it removes the need to generate a proctable per backend, and instead generates one against the frontend.


src/dawn_native/d3d12/DeviceD3D12.cpp
```diff
@@ -39,14 +39,15 @@
 #include "dawn_native/d3d12/SwapChainD3D12.h"
 #include "dawn_native/d3d12/TextureD3D12.h"
 
-namespace dawn_native { namespace d3d12 {
+namespace dawn_native {
+    dawnProcTable GetProcs();
```
SenorBlanco: Nit: forward declaring seems a little bit fragile. Is there a reason we can't #include? The header doesn't even have to be autogen'ed, I think.

Kangz: Agreed, how about I do this in a follow-up patch?

SenorBlanco: Fine by me. (If I thought it was important, I wouldn't have approved. :) )

Kangz:
Thanks!

PR #236, opened 2018-08-02
Initial BUILD.gn files so that it can be used for Skia-Dawn

Kangz:
PTAL

### DIFF
--- a/build_overrides/build.gni
+++ b/build_overrides/build.gni
@@ -1,0 +1,27 @@
+# Copyright 2018 The Dawn Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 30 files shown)


build_overrides/build.gni
```diff
@@ -0,0 +1,28 @@
...
+# Don't use Chromium's third_party/binutils
+linux_use_bundled_binutils_override = false
+
+# In standalone Dawn builds, don't try to use the hermetic install of Xcode
+# that Chromium uses
+use_system_xcode = true
+
+# Android 32-bit non-component, non-clang builds cannot have symbol_level=2
+# due to 4GiB file size limit, see https://crbug.com/648948.
+# Set this flag to true to skip the assertion because we aren't on Android.
```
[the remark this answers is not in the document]
Kangz: Even ANGLE that builds on Android has this set to true. I'll just remove the last line of the comment.

generator/templates/dawn_native/ProcTable.cpp
```diff
@@ -55,6 +55,7 @@ namespace dawn_native {
                         (void) self;
                     {% endif %}
                     bool error = false;
+                    (void) error;
```
kainino0x: Didn't we have an NXT_UNUSED for this?

Kangz: Done

Kangz:
Thanks all. @SenorBlanco, good luck with the rename 😉

PR #239, opened 2018-08-08
BUILD.gn up to tests

Kangz:
PTAL at this new PR!

### DIFF
--- a/scripts/dawn_overrides_with_defaults.gni
+++ b/scripts/dawn_overrides_with_defaults.gni
@@ -1,0 +1,45 @@
+# Copyright 2018 The Dawn Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


SenorBlanco:
Looks good. I think we should land these as separate commits.

Kangz:
Land which parts as separate commits?

SenorBlanco:
I meant to land all four together, but unsquashed. Is that a bad idea? The D3D12 thing seems unrelated to the rest, for example.

Kangz:
Oh, that's what I always do :P

scripts/dawn_overrides_with_defaults.gni
```diff
@@ -0,0 +1,45 @@
...
+
+# This files imports the overrides for Dawn but sets the defaults so that
+# projects including Dawn don't have to set dirs if they happen to use the
+# same.
+# It takes advantage of GN's variable scoping rules to define global variables
+# inside if constructs.
+
+import("//build_overrides/dawn.gni")
+
+if (!defined(dawn_spirv_cross_dir)) {
```
kainino0x: I don't know how gn works, but it _looks_ like what would happen here is:

- importing `build_overrides/dawn.gni` sets `dawn_spirv_cross_dir`
- it's always defined here, so this `if` condition always fails

But I'm probably totally wrong.

Kangz: If you're in a project whose `//build_overrides/dawn.gni` doesn't set the `dawn_spirv_cross_dir` then the `if` is taken. I actually checked it work by commenting out the import (and setting `dawn_standalone` that doesn't have defaults)

Kangz:
Finally figured out you can click "requests" on Travis to see what requests it gets, and that it complained about .travis.yml not parsing.

PR #241, opened 2018-08-13
BUILD.gn: Add samples, cleanups to work from a fresh clone.

Kangz:
PTAL: this makes Dawn compilation work from a fresh checkout using only gclient and gn.


Kangz:
Thanks for the review. @kainino0x TBR to unblock the Windows BUILD.gn PR.

kainino0x:
Does having the .gclient in the root conflict with a chromium build? I don't remember having to bootstrap a .gclient when building ANGLE.

Kangz:
In ANGLE you run `scripts/bootstrap.py` which generates a `.gclient`. Having a .gclient can be an issue when you have a composite .gclient for many project (using fetch chromium, fetch skia, fetch.. for example) where the .gclient file is at the parent directory of the repo. In that situation running gclient will pick-up the checked in .gclient instead of the on at the parent dir.

PR #244, opened 2018-08-16
BUILD.gn: Remove use of exec_script

Kangz:
Chromium's BUILD files try to avoid uses of exec_script when possible
because they slow down every GN invocation. In preparation for building
Dawn inside Chromium, the calls to exec_script for the code generator
are removed.

In GN, the generator now outputs a "JSON tarball", a dictionnary mapping
filenames to content. This allows us to use the "depfile" feature of GN
to avoid the exec_script call to gather the script's inputs.

Outputs of the generator are now listed in the BUILD.gn files. To keep
it in sync with the generator, GN outputs a file containing "expected
outputs" that is checked by the code generator.

Finally the dawn_generator GN template doesn't create a target anymore,
but users are expected to gather outputs using get_target_outputs.


generator/main.py
```diff
@@ -491,26 +460,106 @@ def main():
...
         outputs = set(
             [os.path.abspath(args.output_dir + os.path.sep + render.output) for render in renders]
         )
         outputs = [output.replace('\\', '/') for output in outputs]
-        sys.stdout.write(output_separator.join(outputs))
+        sys.stdout.write(';'.join(outputs))
         return 0
 
-    do_renders(renders, args.template_dir, args.output_dir)
+    # The caller wants to assert that the outputs are what it expect.
```
SenorBlanco: Nit: s/expect/expects/

generator/main.py
```diff
@@ -491,26 +460,106 @@ def main():
...
+    if args.expected_outputs_file != None:
+        with open(args.expected_outputs_file) as f:
+            expected = set([line.strip() for line in f.readlines()])
+
+        actual = set()
+        actual.update([render.output for render in renders])
+
+        if actual != expected:
+            print("Wrong expected outputs, caller expected:\n    " + repr(list(expected)))
+            print("Actually output:\n    " + repr(list(actual)))
```
SenorBlanco: micro-nit: s/Actually/Actual/, to match the noun phrases in the previous sentence (rather than a verb phrase).

generator/main.py
```diff
@@ -491,26 +460,106 @@ def main():
...
+            return 1
+
+    outputs = do_renders(renders, args.template_dir)
+
+    # CMake only, output all the files directly.
```
SenorBlanco: nit: suggest : instead of , (possible comma splice)

generator/main.py
```diff
@@ -491,26 +460,106 @@ def main():
...
+    if args.output_dir != None:
+        output_to_files(outputs, args.output_dir)
+
+    # GN only, output the tarball and its depfile
```
SenorBlanco: nit: suggest : instead of , (possible comma splice)

generator/main.py
Kangz: done

generator/main.py
Kangz: done

generator/main.py
Kangz: done

generator/main.py
Kangz: done

PR #246, opened 2018-08-20
Add correct rpath for dawn samples on linux when using gn build

shaoboyan:
Dawn Samples missing correct rpath settings on linux platform in
BUILD.gn. This failed program to find .so like libdawn_native on linux.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -126,6 +126,20 @@ config("libdawn_public") {
   ]
 }
 
+# Executable needs an rpath to find our shared libraries on OSX and Linux
+config("dawn_shared_library_public") {
+  if (is_mac) {
+    ldflags = [
+      "-rpath",
+      "@executable_path/",
+    ]
+  }
+
+  if (is_linux) {
+    configs = [ "//build/config/gcc:rpath_for_built_shared_libraries" ]
+  }
+}
+
 config("dawn_internal") {
   include_dirs = [ "src" ]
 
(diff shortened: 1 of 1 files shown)


shaoboyan:
This patch fix the issue that sample cannot find .so on linux platform.
@Jiawei-Shao @Kangz PTAL, thx!

shaoboyan:
@Kangz  That's better. Updated the patch and do some rpath clean works too. PTAL.

BUILD.gn
```diff
@@ -150,23 +164,14 @@ config("dawn_internal") {
     defines += [ "DAWN_ENABLE_BACKEND_VULKAN" ]
   }
 
-  configs = [ ":libdawn_public" ]
+  configs = [ ":libdawn_public",
+              ":dawn_shared_library_public", ]
```
Kangz: I was worried that adding it to every internal target would break compilation on OSX because of the added ldflags, but it compiles fine.

BUILD.gn
```diff
@@ -126,6 +126,20 @@ config("libdawn_public") {
   ]
 }
 
+# Executable needs an rpath to find our shared libraries on OSX
```
Kangz: nit: update comment with "OSX and Linux"

BUILD.gn
shaoboyan: Yes, I just  give a try on this and find it works :-)

BUILD.gn
shaoboyan: done.

PR #247, opened 2018-08-20
Workaround a _GLIBCXX_DEBUG bug for std::hash<std::bitset>

Kangz:
This fixes the GN debug build on Linux.

PTAL @SenorBlanco @Jiawei-Shao


Jiawei-Shao:
Now I can compile the debug version of dawn on Linux. Thanks for your fix!

Kangz:
Thanks for testing it works for you. I should have done this workaround a long time ago.

PR #250, opened 2018-08-20
Remove BindGroupBuilder::SetUsage

Kangz:
BindGroup usage isn't something that's part of WebGPU's sketch.idl and
it might never exist. Remove it to simplify the migration of bindgroup
to descriptor.

PTAL @SenorBlanco @kainino0x

### DIFF
--- a/dawn.json
+++ b/dawn.json
@@ -34,12 +34,6 @@
                 ]
             },
             {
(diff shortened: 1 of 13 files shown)


Kangz:
Thanks for the review (for this and other PRs)

PR #251, opened 2018-08-22
Descriptorize Buffer

Kangz:
PTAL, this was more painful that it should have been but it's finally done!


Kangz:
@kainino0x @SenorBlanco PTAL

dawn.json
```diff
@@ -537,7 +540,14 @@
                 "returns": "blend state builder"
             },
             {
-                "name": "create buffer builder",
+                "name": "create buffer",
+                "returns": "buffer",
+                "args": [
+                    {"name": "descriptor", "type": "buffer descriptor", "annotation": "const*"}
```
SenorBlanco: I'm probably forgetting something, but for objects this small (two params), could we just pass them as args instead of in a descriptor? Or is it for future expansion?

Kangz: Like you guessed the descriptor gives us extensibility.

PR #253, opened 2018-08-23
cr-buildbucket.cfg: Fix missing clang and no_clang mixins

Kangz:
PTAL @SenorBlanco 
I already landed this on Gerrit to see if it would fix luci-config (it does!) and I'll force push again once this commit is landed on Github.


Kangz:
Thanks!

PR #254, opened 2018-08-23
 Fixes for cq.cfg

Kangz:
A fun bug happened because our CQ config landed both global/ and
branch/ in the same commit. LUCI tries to discover where cq.cfg with
refs.cfg, however the processing of cq.cfg is done before refs.cfg. This
means that when everything was landed at once, the processing of cq.cfg
didn't where to look, and then refs.cfg was processed, but too late.

So this whitespace change will dirty cq.cfg and cause LUCI to discover
it.

PTAL @SenorBlanco

### DIFF
--- a/infra/config/branch/cq.cfg
+++ b/infra/config/branch/cq.cfg
@@ -2,9 +2,11 @@
 # documentation of this file format.
 
 version: 1
-cq_name: "dawn"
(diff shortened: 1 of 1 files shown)


Kangz:
Thanks!

PR #256, opened 2018-08-23
Fix VkNonDispatchableHandle's alignment on 32bit platforms

Kangz:
On 32bit platforms, even if it contains a uint64_t the handle wrapper
will align to 4 bytes because the assembly load operations can't take
advantage of 64bit aligned loads.

Fix this by forcing the alignment to the alignment of what would be the
native Vulkan handle type.

PTAL @SenorBlanco @kainino0x

### DIFF
--- a/src/common/vulkan_platform.h
+++ b/src/common/vulkan_platform.h
@@ -32,9 +32,21 @@
 // redefined to be nullptr). This keeps the type-safety of having the handles be different types
 // (like vulkan.h on 64 bit) but makes sure the types are different on 32 bit architectures.
 
+// Force the handle type to have the same alignment as what would have been the Vulkan
+// non-dispatchable handle type.
+#if defined(DAWN_PLATFORM_64_BIT)
+// In 64 bit handles are pointers to some structure, we just declare one inline here.
+using NativeVulkanHandleType = struct VkSomeHandle*;
+#elif defined(DAWN_PLATFORM_32_BIT)
+using NativeVulkanHandleType = uint64_t;
+#    define ALIGNAS_VULKAN_HANDLE alignas(alignof(uint64_t))
+#else
+#    error "Unsupported platform"
+#endif
+
 // Simple handle types that supports "nullptr_t" as a 0 value.
 template <typename Tag, typename HandleType>
-class VkNonDispatchableHandle {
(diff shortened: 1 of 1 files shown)


src/common/vulkan_platform.h
```diff
@@ -32,9 +32,19 @@
 // redefined to be nullptr). This keeps the type-safety of having the handles be different types
 // (like vulkan.h on 64 bit) but makes sure the types are different on 32 bit architectures.
 
+// Force the handle type to have the same alignment as what would have been the Vulkan
+// non-dispatchable handle type.
+#if DAWN_PLATFORM_64_BIT
+#    define ALIGNAS_VULKAN_HANDLE alignas(alignof(struct VkSomeHandle*))
```
kainino0x: Kind of unclear where the definition of VkSomeHandle is coming from (after a minute I realized this IS the definition) - a comment could be nice. (Or maybe use a real struct that's already in scope?)

Kangz: Added a comment, also changed code in the #if to declare a type. This way the alignas(alignof()) is not duplicated.

src/common/vulkan_platform.h
```diff
@@ -32,9 +32,21 @@
 // redefined to be nullptr). This keeps the type-safety of having the handles be different types
 // (like vulkan.h on 64 bit) but makes sure the types are different on 32 bit architectures.
 
+// Force the handle type to have the same alignment as what would have been the Vulkan
+// non-dispatchable handle type.
+#if defined(DAWN_PLATFORM_64_BIT)
+// In 64 bit handles are pointers to some structure, we just declare one inline here.
+using NativeVulkanHandleType = struct VkSomeHandle*;
+#elif defined(DAWN_PLATFORM_32_BIT)
+using NativeVulkanHandleType = uint64_t;
```
SenorBlanco: Maybe a dumb question, but why not just use uint64_t in all cases? In 64-bit, we know a pointer is 64-bit, in 32-bit we're forcing it to 64 bits.

Kangz: It's because for type-safety the Vulkan headers use `uint64_t` on 32bit but `struct VkType_T*` on 64bit that can't cast to one another. So we have to do the same in case the size and alignment of pointers isn't the same on 64bit.

PR #257, opened 2018-08-24
Implement copying between a buffer and a texture 2D array on Vulkan

Jiawei-Shao:
This patch implements the creation of a 2D array texture and
copying data between a buffer and a layer of a 2D array texture
on Vulkan back-ends. A new parameter "slice" is added to both
CopyBufferToTexture and CopyTextureToBuffer to specify the
slice of the 2D array texture in this copy operation.

TEST=dawn_end2end_tests

### DIFF
--- a/src/dawn_native/CommandBuffer.cpp
+++ b/src/dawn_native/CommandBuffer.cpp
@@ -46,15 +46,20 @@ namespace dawn_native {
                 uint64_t(location.y) + uint64_t(location.height) >
                     (static_cast<uint64_t>(texture->GetHeight()) >> level) ||
                 uint64_t(location.z) + uint64_t(location.depth) >
-                    (static_cast<uint64_t>(texture->GetDepth()))){
+                    (static_cast<uint64_t>(texture->GetDepth()))) {
                 DAWN_RETURN_ERROR("Copy would touch outside of the texture");
             }
 
-            // TODO(jiawei.shao@intel.com): support 1D and 3D textures
-            ASSERT(texture->GetDimension() == dawn::TextureDimension::e2D);
-            if (location.depth != 1u)
-            {
-                DAWN_RETURN_ERROR("No support for copying to multiple 2D array layers for now");
+            switch (texture->GetDimension()) {
+                case dawn::TextureDimension::e2D:
+                    if (location.depth != 1u) {
+                        DAWN_RETURN_ERROR("No support for copying to multiple 2D array layers for now");
+                    }
+                    break;
+
+                // TODO(jiawei.shao@intel.com): support 1D and 3D textures
+                default:
+                    UNREACHABLE();
             }
 
             return {};
--- a/src/dawn_native/CommandBuffer.cpp
+++ b/src/dawn_native/CommandBuffer.cpp
@@ -53,7 +53,8 @@ namespace dawn_native {
             switch (texture->GetDimension()) {
                 case dawn::TextureDimension::e2D:
                     if (location.depth != 1u) {
-                        DAWN_RETURN_ERROR("No support for copying to multiple 2D array layers for now");
+                        DAWN_RETURN_ERROR(
+                            "No support for copying to multiple 2D array layers for now");
                     }
                     break;
 
--- a/src/dawn_native/CommandBuffer.cpp
+++ b/src/dawn_native/CommandBuffer.cpp
@@ -38,29 +38,24 @@ namespace dawn_native {
                 DAWN_RETURN_ERROR("Copy mip-level out of range");
             }
 
+            if (location.slice >= texture->GetArrayLayers()) {
+                DAWN_RETURN_ERROR("Copy array-layer out of range");
+            }
+
             // All texture dimensions are in uint32_t so by doing checks in uint64_t we avoid
             // overflows.
             uint64_t level = location.level;
             if (uint64_t(location.x) + uint64_t(location.width) >
                     (static_cast<uint64_t>(texture->GetWidth()) >> level) ||
                 uint64_t(location.y) + uint64_t(location.height) >
-                    (static_cast<uint64_t>(texture->GetHeight()) >> level) ||
-                uint64_t(location.z) + uint64_t(location.depth) >
-                    (static_cast<uint64_t>(texture->GetDepth()))) {
+                    (static_cast<uint64_t>(texture->GetHeight()) >> level)) {
                 DAWN_RETURN_ERROR("Copy would touch outside of the texture");
             }
 
-            switch (texture->GetDimension()) {
-                case dawn::TextureDimension::e2D:
-                    if (location.depth != 1u) {
-                        DAWN_RETURN_ERROR(
-                            "No support for copying to multiple 2D array layers for now");
-                    }
-                    break;
-
-                // TODO(jiawei.shao@intel.com): support 1D and 3D textures
-                default:
-                    UNREACHABLE();
+            // TODO(cwallez@chromium.org): Check the depth bound differently for 2D arrays and 3D
+            // textures
+            if (location.z != 0 || location.depth != 1) {
+                DAWN_RETURN_ERROR("No support for z != 0 and depth != 1 for now");
             }
 
             return {};
@@ -594,7 +589,8 @@ namespace dawn_native {
                                                    uint32_t width,
                                                    uint32_t height,
                                                    uint32_t depth,
-                                                   uint32_t level) {
+                                                   uint32_t level,
+                                                   uint32_t slice) {
         if (rowPitch == 0) {
             rowPitch = ComputeDefaultRowPitch(texture, width);
         }
@@ -611,6 +607,7 @@ namespace dawn_native {
         copy->destination.height = height;
         copy->destination.depth = depth;
         copy->destination.level = level;
+        copy->destination.slice = slice;
         copy->rowPitch = rowPitch;
     }
 
@@ -622,6 +619,7 @@ namespace dawn_native {
                                                    uint32_t height,
                                                    uint32_t depth,
                                                    uint32_t level,
+                                                   uint32_t slice,
                                                    BufferBase* buffer,
                                                    uint32_t bufferOffset,
                                                    uint32_t rowPitch) {
@@ -639,6 +637,7 @@ namespace dawn_native {
         copy->source.height = height;
         copy->source.depth = depth;
         copy->source.level = level;
+        copy->source.slice = slice;
         copy->destination.buffer = buffer;
         copy->destination.offset = bufferOffset;
         copy->rowPitch = rowPitch;
--- a/src/dawn_native/CommandBuffer.cpp
+++ b/src/dawn_native/CommandBuffer.cpp
@@ -44,14 +44,17 @@ namespace dawn_native {
             if (uint64_t(location.x) + uint64_t(location.width) >
                     (static_cast<uint64_t>(texture->GetWidth()) >> level) ||
                 uint64_t(location.y) + uint64_t(location.height) >
-                    (static_cast<uint64_t>(texture->GetHeight()) >> level)) {
+                    (static_cast<uint64_t>(texture->GetHeight()) >> level) ||
+                uint64_t(location.z) + uint64_t(location.depth) >
+                    (static_cast<uint64_t>(texture->GetDepth()))){
                 DAWN_RETURN_ERROR("Copy would touch outside of the texture");
             }
 
-            // TODO(cwallez@chromium.org): Check the depth bound differently for 2D arrays and 3D
-            // textures
-            if (location.z != 0 || location.depth != 1) {
-                DAWN_RETURN_ERROR("No support for z != 0 and depth != 1 for now");
+            // TODO(jiawei.shao@intel.com): support 1D and 3D textures
+            ASSERT(texture->GetDimension() == dawn::TextureDimension::e2D);
+            if (location.depth != 1u)
+            {
+                DAWN_RETURN_ERROR("No support for copying to multiple 2D array layers for now");
             }
 
             return {};
--- a/src/dawn_native/vulkan/CommandBufferVk.cpp
+++ b/src/dawn_native/vulkan/CommandBufferVk.cpp
@@ -53,16 +53,19 @@ namespace dawn_native { namespace vulkan {
 
             region.imageSubresource.aspectMask = texture->GetVkAspectMask();
             region.imageSubresource.mipLevel = textureLocation.level;
-            region.imageSubresource.baseArrayLayer = 0;
+
+            // TODO(jiawei.shao@intel.com): support 1D and 3D textures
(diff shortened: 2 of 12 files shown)


Jiawei-Shao:
@shaoboyan @gyagp

src/dawn_native/vulkan/CommandBufferVk.cpp
```diff
@@ -53,16 +53,19 @@ namespace dawn_native { namespace vulkan {
 
             region.imageSubresource.aspectMask = texture->GetVkAspectMask();
             region.imageSubresource.mipLevel = textureLocation.level;
-            region.imageSubresource.baseArrayLayer = 0;
+
+            // TODO(jiawei.shao@intel.com): support 1D and 3D textures
+            ASSERT(texture->GetDimension() == dawn::TextureDimension::e2D);
```
kainino0x: drive-by: I'd prefer if this section looked a little more like we might expect in final code, e.g.:
```cpp
if (texture->GetDimension() == dawn::TextureDimension::e2D) {
  // set the appropriate fields of `region`
} else {
  ASSERT(false); // or something; I don't think we have a NOTIMPLEMENTED but you could add one
}```

Comment (kainino0x) [reply src/dawn_native/vulkan/CommandBufferVk.cpp]: (Same comment for other similar places in this pull request)

File: src/dawn_native/CommandBuffer.cpp [hunk 2]
```diff
@@ -44,14 +44,23 @@ namespace dawn_native {
...
+                    (static_cast<uint64_t>(texture->GetDepth()))) {
                 DAWN_RETURN_ERROR("Copy would touch outside of the texture");
             }
 
-            // TODO(cwallez@chromium.org): Check the depth bound differently for 2D arrays and 3D
-            // textures
-            if (location.z != 0 || location.depth != 1) {
-                DAWN_RETURN_ERROR("No support for z != 0 and depth != 1 for now");
+            switch (texture->GetDimension()) {
+                case dawn::TextureDimension::e2D:
```
Kangz: After [1] this check should be replaced by a bounds check on array slices.

src/dawn_native/vulkan/TextureVk.cpp
```diff
@@ -258,6 +256,17 @@ namespace dawn_native { namespace vulkan {
         createInfo.pQueueFamilyIndices = nullptr;
         createInfo.initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
 
+        switch (GetDimension()) {
+            case dawn::TextureDimension::e2D:
+                createInfo.extent = VkExtent3D{GetWidth(), GetHeight(), 1};
+                createInfo.arrayLayers = GetDepth();
```
Kangz: [1]: I don't think that depth will be for both 3D texture depth and the 2D array size. Vulkan and Metal have it separate while only D3D12 has the combined `DepthOrArraySize` in the texture descriptors. Can a new "SetArraySlice" be added to the TextureBuilder, or even better could the texture be descriptorized? It is something we need to do to get closer to WebGPU anyway.

src/dawn_native/vulkan/CommandBufferVk.cpp
```diff
@@ -53,16 +53,23 @@ namespace dawn_native { namespace vulkan {
...
-            region.imageOffset.x = textureLocation.x;
-            region.imageOffset.y = textureLocation.y;
-            region.imageOffset.z = textureLocation.z;
-
-            region.imageExtent.width = textureLocation.width;
-            region.imageExtent.height = textureLocation.height;
-            region.imageExtent.depth = textureLocation.depth;
+            switch (texture->GetDimension()) {
+                case dawn::TextureDimension::e2D:
+                    region.imageSubresource.baseArrayLayer = textureLocation.z;
```
Kangz: Like in [1] this should probably be a separate argument, see the proposal for texture copies [here](https://github.com/gpuweb/gpuweb/issues/69).

src/dawn_native/CommandBuffer.cpp
```diff
@@ -44,14 +44,23 @@ namespace dawn_native {
             if (uint64_t(location.x) + uint64_t(location.width) >
                     (static_cast<uint64_t>(texture->GetWidth()) >> level) ||
                 uint64_t(location.y) + uint64_t(location.height) >
-                    (static_cast<uint64_t>(texture->GetHeight()) >> level)) {
+                    (static_cast<uint64_t>(texture->GetHeight()) >> level) ||
+                uint64_t(location.z) + uint64_t(location.depth) >
```
Kangz: After [1] an additional check for array slice will be required too.

Kangz: New validation should also be covered by a [CopyCommandValidationTest](https://github.com/google/nxt-standalone/blob/master/src/tests/unittests/validation/CopyCommandsValidationTests.cpp).

src/tests/end2end/CopyTests.cpp
```diff
@@ -84,18 +85,23 @@ class CopyTests_T2B : public CopyTests {
...
             uint32_t texelsPerRow = rowPitch / kBytesPerTexel;
-            uint32_t texelCount = texelsPerRow * (height - 1) + width;
+            uint32_t texelCountPerLayer = texelsPerRow * (height - 1) + width;
 
             // Create an upload buffer and use it to populate the `level` mip of the texture
-            std::vector<RGBA8> textureData(texelCount);
-            FillTextureData(width, height, rowPitch / kBytesPerTexel, textureData.data());
-            dawn::Buffer uploadBuffer = utils::CreateBufferFromData(device, textureData.data(), static_cast<uint32_t>(sizeof(RGBA8) * textureData.size()), dawn::BufferUsageBit::TransferSrc);
+            std::vector<std::vector<RGBA8>> textureArrayData(textureSpec.arrayLayers);
+            std::vector<dawn::CommandBuffer> commands(textureSpec.arrayLayers);
```
Kangz: We can keep only one command buffer and have multiple copies in it.

src/tests/end2end/CopyTests.cpp
```diff
@@ -84,18 +85,23 @@ class CopyTests_T2B : public CopyTests {
             uint32_t height = textureSpec.height >> textureSpec.level;
             uint32_t rowPitch = Align(kBytesPerTexel * width, kTextureRowPitchAlignment);
             uint32_t texelsPerRow = rowPitch / kBytesPerTexel;
-            uint32_t texelCount = texelsPerRow * (height - 1) + width;
+            uint32_t texelCountPerLayer = texelsPerRow * (height - 1) + width;
 
             // Create an upload buffer and use it to populate the `level` mip of the texture
```
Kangz: nit: fix comment to refer to multiple buffers.

Alternatively we could have a single buffer with all the layers and then do copies starting at an offset that depends on the layer.

src/tests/end2end/CopyTests.cpp
```diff
@@ -106,32 +112,31 @@ class CopyTests_T2B : public CopyTests {
...
-
-            EXPECT_BUFFER_U32_RANGE_EQ(reinterpret_cast<const uint32_t*>(expected.data()), buffer, bufferSpec.offset, static_cast<uint32_t>(expected.size())) <<
-                "Texture to Buffer copy failed copying region [(" << textureSpec.x << ", " << textureSpec.y << "), (" << textureSpec.x + textureSpec.copyWidth << ", " << textureSpec.y + textureSpec.copyHeight <<
-                ")) from " << textureSpec.width << " x " << textureSpec.height << " texture at mip level " << textureSpec.level <<
-                " to " << bufferSpec.size << "-byte buffer with offset " << bufferSpec.offset << " and row pitch " << bufferSpec.rowPitch << std::endl;
+            for (uint32_t layer = 0; layer < textureSpec.arrayLayers; ++layer) {
+                buffer.SetSubData(0, static_cast<uint32_t>(emptyData.size() * sizeof(RGBA8)), reinterpret_cast<const uint8_t*>(emptyData.data()));
+
+                // Copy the region [(`x`, `y`), (`x + copyWidth, `y + copyWidth`)] from the `level` mip into the buffer at the specified `offset` and `rowPitch`
+                dawn::CommandBuffer command = device.CreateCommandBufferBuilder()
```
Kangz: This can still be in the same command buffer as the upload.

src/tests/end2end/CopyTests.cpp
```diff
@@ -74,7 +75,7 @@ class CopyTests_T2B : public CopyTests {
             // Create a texture that is `width` x `height` with (`level` + 1) mip levels.
```
Kangz: nit: update the comment

src/dawn_native/vulkan/TextureVk.cpp
Jiawei-Shao: Got it. I will descriptorize texture first, then support 2D array textures. The implementation of "BufferDescriptor" should be a good reference for me.

src/dawn_native/vulkan/CommandBufferVk.cpp
Jiawei-Shao: I once thought as there were no "3D array" textures, maybe we could avoid using one more variable here. Sorry that I failed to catch this key point in this proposal.

src/dawn_native/vulkan/TextureVk.cpp
Kangz: Thanks!

src/dawn_native/vulkan/CommandBufferVk.cpp
Kangz: No worries, I don't think there are 3D array textures in any API (and actually you might be able to convert 2D array to a 3D texture view in some conditions) so things are a bit confusing.

src/dawn_native/vulkan/TextureVk.cpp
Jiawei-Shao: Done

src/dawn_native/CommandBuffer.cpp
Jiawei-Shao: Done

src/tests/end2end/CopyTests.cpp
Jiawei-Shao: Done

src/tests/end2end/CopyTests.cpp
Jiawei-Shao: Done

src/tests/end2end/CopyTests.cpp
```diff
@@ -355,6 +368,36 @@ TEST_P(CopyTests_T2B, RowPitchUnaligned) {
     }
 }
 
+// Test that copying regions of each texture 2D array layer works
+TEST_P(CopyTests_T2B, Texture2DArrayRegion)
+{
+    // TODO(jiawei.shao@intel.com): support 2D array texture on OpenGL, D3D12 and Metal.
```
Kangz: In a follow-up patch could we have a DAWN_SKIP_TEST_IF like ANGLE_SKIP_TEST_IF?

src/tests/end2end/CopyTests.cpp
```diff
@@ -87,51 +88,63 @@ class CopyTests_T2B : public CopyTests {
...
-            dawn::CommandBuffer commands[2];
-
-            commands[0] = device.CreateCommandBufferBuilder()
-                .CopyBufferToTexture(uploadBuffer, 0, rowPitch, texture, 0, 0, 0, width, height, 1, textureSpec.level)
-                .GetResult();
+            uint32_t texelCountPerLayer = texelsPerRow * (height - 1) + width;
+
+            // Create an upload buffer and use it to populate all slices of the `level` mip of the texture
+            std::vector<std::vector<RGBA8>> textureArrayData(textureSpec.arrayLayer);
+            std::vector<dawn::CommandBuffer> commands(textureSpec.arrayLayer * 2);
```
Kangz: We can keep only one command buffer and have multiple copies in it.

src/tests/end2end/CopyTests.cpp
```diff
@@ -87,51 +88,63 @@ class CopyTests_T2B : public CopyTests {
...
-            dawn::Buffer uploadBuffer = utils::CreateBufferFromData(device, textureData.data(), static_cast<uint32_t>(sizeof(RGBA8) * textureData.size()), dawn::BufferUsageBit::TransferSrc);
-
-            dawn::CommandBuffer commands[2];
-
-            commands[0] = device.CreateCommandBufferBuilder()
-                .CopyBufferToTexture(uploadBuffer, 0, rowPitch, texture, 0, 0, 0, width, height, 1, textureSpec.level)
-                .GetResult();
+            uint32_t texelCountPerLayer = texelsPerRow * (height - 1) + width;
+
+            // Create an upload buffer and use it to populate all slices of the `level` mip of the texture
```
Kangz: nit: fixup comments

src/tests/end2end/CopyTests.cpp
```diff
@@ -87,51 +88,63 @@ class CopyTests_T2B : public CopyTests {
...
             buffer.SetSubData(0, static_cast<uint32_t>(emptyData.size() * sizeof(RGBA8)), reinterpret_cast<const uint8_t*>(emptyData.data()));
 
-            // Copy the region [(`x`, `y`), (`x + copyWidth, `y + copyWidth`)] from the `level` mip into the buffer at the specified `offset` and `rowPitch`
-            commands[1] = device.CreateCommandBufferBuilder()
-                .CopyTextureToBuffer(texture, textureSpec.x, textureSpec.y, 0, textureSpec.copyWidth, textureSpec.copyHeight, 1, textureSpec.level, buffer, bufferSpec.offset, bufferSpec.rowPitch)
-                .GetResult();
+            uint32_t bufferOffset = bufferSpec.offset;
+            for (uint32_t slice = 0; slice < textureSpec.arrayLayer; ++slice) {
+                // Copy the region [(`x`, `y`), (`x + copyWidth, `y + copyWidth`)] from the `level` mip into the buffer at `offset + bufferSpec.size * slice` and `rowPitch`
+                commands[textureSpec.arrayLayer + slice] = device.CreateCommandBufferBuilder()
```
Kangz: this can still be in the same single command buffer as the uploads.

src/tests/end2end/CopyTests.cpp
Jiawei-Shao: Ack

src/tests/end2end/CopyTests.cpp
Jiawei-Shao: Done

src/tests/end2end/CopyTests.cpp
Jiawei-Shao: Done

src/tests/end2end/CopyTests.cpp
Jiawei-Shao: Done

Kangz:
LGTM, thanks for the updates!

PR #258, opened 2018-08-24
Make all CI bots scheduled by the gittiles trigger

Kangz:
PTAL @SenorBlanco


Kangz:
Landing this even is AppVeyor didn't finish because the change doesn't impact the testing there.

PR #259, opened 2018-08-24
Roll SPIRV-Cross and use Chromium git mirrors in DEPS

Kangz:
PTAL @SenorBlanco @kainino0x

### DIFF
--- a/src/dawn_native/ShaderModule.cpp
+++ b/src/dawn_native/ShaderModule.cpp
@@ -71,8 +71,9 @@ namespace dawn_native {
             ASSERT(blockType.basetype == spirv_cross::SPIRType::Struct);
 
             for (uint32_t i = 0; i < blockType.member_types.size(); i++) {
-                ASSERT(compiler.get_member_decoration_mask(blockType.self, i) &
-                       1ull << spv::DecorationOffset);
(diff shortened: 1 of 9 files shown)


src/dawn_native/ShaderModule.cpp
```diff
@@ -114,12 +115,11 @@ namespace dawn_native {
         auto ExtractResourcesBinding = [this](const std::vector<spirv_cross::Resource>& resources,
                                               const spirv_cross::Compiler& compiler,
                                               dawn::BindingType bindingType) {
-            constexpr uint64_t requiredBindingDecorationMask =
-                (1ull << spv::DecorationBinding) | (1ull << spv::DecorationDescriptorSet);
-
             for (const auto& resource : resources) {
-                ASSERT((compiler.get_decoration_mask(resource.id) &
-                        requiredBindingDecorationMask) == requiredBindingDecorationMask);
+                ASSERT(compiler.get_decoration_bitset(resource.id).get(spv::DecorationBinding));
```
SenorBlanco: I assume these changes are necessary to roll spirv-cross, not unrelated?

Kangz: Yep, the functions we were using before were marked as deprecated and caused warnings hence errors.

PR #261, opened 2018-08-27
Descriptorize Texture

Jiawei-Shao:
This patch introduces texture descriptor for texture creation instead
of texture builders. Like BufferBuilder, we keep a dummy TextureBuilder
for the builder error callback tests of wire.

This patch also adds "arrayLayer" to texture descriptor based on
https://github.com/gpuweb/gpuweb/issues/68.


Jiawei-Shao:
@kangz @kainino0x PTAL, thanks!

dawn.json
```diff
@@ -997,6 +1004,7 @@
         ]
     },
     "texture builder": {
+        "_comment": "This builder is kept for testing only",
```
Kangz: After your patch there is no more mention of the texture builder so this isn't needed. We kept it for buffers because it was the builder being used to test the "wire" implementation of the builder error callback. This builder isn't used like that so it can be safely removed.

dawn.json
```diff
@@ -1028,15 +1036,23 @@
...
+    "texture descriptor": {
+        "category": "structure",
+        "extensible": true,
+        "members": [
+            {"name": "usage", "type": "texture usage bit"},
+            {"name": "dimension", "type": "texture dimension"},
+            {"name": "width", "type": "uint32_t"},
+            {"name": "height", "type": "uint32_t"},
+            {"name": "depth", "type": "uint32_t"},
+            {"name": "arrayLayer", "type": "uint32_t"},
```
Kangz: `arrayLayer` LGTM, we might need to rename it when WebGPU decides what name it wants to use.

dawn.json
```diff
@@ -1028,15 +1036,23 @@
...
             }
         ]
     },
+    "texture descriptor": {
+        "category": "structure",
+        "extensible": true,
+        "members": [
+            {"name": "usage", "type": "texture usage bit"},
+            {"name": "dimension", "type": "texture dimension"},
+            {"name": "width", "type": "uint32_t"},
```
Kangz: FYI: at some point `width`, `height` and `depth` will probably become a nested `dawn::Extent3D` structure like in WebGPU. We can do it when the WebGPU API crystallizes a bit more though.

dawn.json
Kangz: Also FYI, there will be default values for these members so we don't have to type `arrayLayers = 1` every time. Likewise this will be a bit later when WebGPU crystallizes.

src/dawn_native/Device.cpp
```diff
@@ -178,8 +178,14 @@ namespace dawn_native {
     SwapChainBuilder* DeviceBase::CreateSwapChainBuilder() {
         return new SwapChainBuilder(this);
     }
-    TextureBuilder* DeviceBase::CreateTextureBuilder() {
-        return new TextureBuilder(this);
+    TextureBase* DeviceBase::CreateTexture(const TextureDescriptor* descriptor) {
+        TextureBase* result = nullptr;
+
+        if (ConsumedError(CreateTextureInternal(&result, descriptor)))
+        {
```
Kangz: nit: contrary to ANGLE we put braces at the end of the line. We have a clang format file, but it doesn't apply to all files yet (tests get really badly formatted because clang-format can't handle the "builder pattern"). Normally the formatting errors will be showed in the Travis build log with a diff you can apply to fix them. The only issue is that for now the LLVM apt repo is broken so the Travis build fails too early.

src/dawn_native/Device.cpp
```diff
@@ -218,6 +224,13 @@ namespace dawn_native {
         return {};
     }
 
+    MaybeError DeviceBase::CreateTextureInternal(TextureBase** result,
```
Kangz: ultra-nit: can this be sorted among the CreateFooInternal function?

src/dawn_native/Device.h
```diff
@@ -105,10 +104,15 @@ namespace dawn_native {
...
 
+        TextureBuilder* CreateTextureBuilderForTesting() {
+            return nullptr;
+        }
+
       private:
         virtual ResultOrError<BindGroupLayoutBase*> CreateBindGroupLayoutImpl(
             const BindGroupLayoutDescriptor* descriptor) = 0;
         virtual ResultOrError<BufferBase*> CreateBufferImpl(const BufferDescriptor* descriptor) = 0;
+        virtual ResultOrError<TextureBase*> CreateTextureImpl(const TextureDescriptor* descriptor) = 0;
```
Kangz: ultra-nit: sort.

src/dawn_native/Device.h
```diff
@@ -120,6 +124,7 @@ namespace dawn_native {
         MaybeError CreateBindGroupLayoutInternal(BindGroupLayoutBase** result,
                                                  const BindGroupLayoutDescriptor* descriptor);
         MaybeError CreateBufferInternal(BufferBase** result, const BufferDescriptor* descriptor);
+        MaybeError CreateTextureInternal(TextureBase** result, const TextureDescriptor* descriptor);
```
Kangz: ultra-nit: sort

src/dawn_native/Texture.h
```diff
@@ -111,6 +89,30 @@ namespace dawn_native {
         Ref<TextureBase> mTexture;
     };
 
+    // This builder class is kept around purely for testing but should not be used.
```
Kangz: It can even be removed!

src/dawn_native/d3d12/DeviceD3D12.h
```diff
@@ -78,6 +77,7 @@ namespace dawn_native { namespace d3d12 {
         ResultOrError<BindGroupLayoutBase*> CreateBindGroupLayoutImpl(
             const BindGroupLayoutDescriptor* descriptor) override;
         ResultOrError<BufferBase*> CreateBufferImpl(const BufferDescriptor* descriptor) override;
+        ResultOrError<TextureBase*> CreateTextureImpl(const TextureDescriptor* descriptor) override;
```
Kangz: nit: sort, here and other DeviceXXX and in NullBackend too

src/dawn_native/d3d12/TextureD3D12.cpp
```diff
@@ -107,8 +107,8 @@ namespace dawn_native { namespace d3d12 {
         }
     }
 
-    Texture::Texture(TextureBuilder* builder)
-        : TextureBase(builder), mDevice(ToBackend(builder->GetDevice())) {
+    Texture::Texture(Device* device, const TextureDescriptor* descriptor)
+        : TextureBase(device, descriptor), mDevice(device) {
```
Kangz: TextureD3D12 shouldn't have an mDevice but instead use ToBackend(GetDevice()). This need to be addressed eventually but it isn't necessary for this patch.

Kangz:
Thanks it looks great, there's some formatting errors but Travis is failing right now, I'll try to fix it so you can copy paste the diff to apply.

Kangz:
It should be fixed now but I think you need to rebase on top of master and force-push to your branch for Travis to see the updated .travis.yml

src/dawn_native/Device.cpp
Jiawei-Shao: Done

src/dawn_native/Texture.h
Jiawei-Shao: Done

src/dawn_native/Device.h
Jiawei-Shao: Done

src/dawn_native/Device.h
Jiawei-Shao: Done

src/dawn_native/d3d12/DeviceD3D12.h
Jiawei-Shao: Done

src/dawn_native/d3d12/TextureD3D12.cpp
Jiawei-Shao: Done

Kangz:
LGTM thanks for the updates!

PR #264, opened 2018-08-29
GN: Fix MSVC compilation

Kangz:
third_party/BUILD.gn was adding compile flags to suppress failures that
were GCC/Clang specific and caused MSVC to error out. Replace them by
suppression using MSVC's flags.

PTAL @kainino0x @SenorBlanco I verified it passes the MSVC bots in the tryjobs for this CL: https://dawn-review.googlesource.com/c/dawn/+/1220


Kangz:
Thank you!

PR #268, opened 2018-08-31
Support 2D array texture copying on D3D12, Metal and OpenGL

Jiawei-Shao:
This patch implements the creation of 2D array textures and copying
between a buffer and a layer of a 2D array texture on D3D12, Metal
and OpenGL back-ends.

This patch also fixes the barrier in vulkan::Texture::TransitionUsageNow
to make the barrier cover all the slices in a 2D array texture.

TEST=dawn_end2end_tests

### DIFF
--- a/src/dawn_native/opengl/CommandBufferGL.cpp
+++ b/src/dawn_native/opengl/CommandBufferGL.cpp
@@ -331,13 +331,28 @@ namespace dawn_native { namespace opengl {
                     glActiveTexture(GL_TEXTURE0);
                     glBindTexture(target, texture->GetHandle());
 
-                    ASSERT(texture->GetDimension() == dawn::TextureDimension::e2D);
                     glPixelStorei(GL_UNPACK_ROW_LENGTH,
                                   copy->rowPitch / TextureFormatPixelSize(texture->GetFormat()));
-                    glTexSubImage2D(target, dst.level, dst.x, dst.y, dst.width, dst.height,
+                    switch (texture->GetDimension()) {
+                        case dawn::TextureDimension::e2D:
+                            if (texture->GetArrayLayers() > 1) {
+                                glTexSubImage3D(
+                                    target, dst.level, dst.x, dst.y, dst.slice, dst.width,
+                                    dst.height, 1, format.format, format.type,
+                                    reinterpret_cast<void*>(static_cast<uintptr_t>(src.offset)));
+                            } else {
+                                glTexSubImage2D(
+                                    target, dst.level, dst.x, dst.y, dst.width, dst.height,
                                     format.format, format.type,
                                     reinterpret_cast<void*>(static_cast<uintptr_t>(src.offset)));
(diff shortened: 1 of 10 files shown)


Jiawei-Shao:
@Kangz @kainino0x PTAL, thanks!

Jiawei-Shao:
Oh I missed this point. Thanks to point out this!

I see in d3d12::Texture::TransitionUsageNow() we are using D3D12_RESOURCE_BARRIER_ALL_SUBRESOURCES as barrier.Transition.Subresource, so I think I don't need to modify the D3D12 part.

Kangz:
Thanks for fixing it, still LGTM. Let's wait for @kainino0x's review.

src/dawn_native/opengl/CommandBufferGL.cpp
```diff
@@ -331,13 +331,28 @@ namespace dawn_native { namespace opengl {
...
-                    ASSERT(texture->GetDimension() == dawn::TextureDimension::e2D);
                     glPixelStorei(GL_UNPACK_ROW_LENGTH,
                                   copy->rowPitch / TextureFormatPixelSize(texture->GetFormat()));
-                    glTexSubImage2D(target, dst.level, dst.x, dst.y, dst.width, dst.height,
+                    switch (texture->GetDimension()) {
+                        case dawn::TextureDimension::e2D:
+                            if (texture->GetArrayLayers() > 1) {
+                                glTexSubImage3D(
+                                    target, dst.level, dst.x, dst.y, dst.slice, dst.width,
+                                    dst.height, 1, format.format, format.type,
```
kainino0x: Shouldn't we be using dst.depth here? Might be tricky to implement the same semantics in OpenGL, though, I'm not sure.

src/dawn_native/opengl/CommandBufferGL.cpp
```diff
@@ -348,18 +363,31 @@ namespace dawn_native { namespace opengl {
...
                     glBindFramebuffer(GL_READ_FRAMEBUFFER, readFBO);
-
-                    glFramebufferTexture2D(GL_READ_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D,
-                                           texture->GetHandle(), src.level);
+                    switch (texture->GetDimension()) {
+                        case dawn::TextureDimension::e2D:
+                            if (texture->GetArrayLayers() > 1) {
+                                glFramebufferTextureLayer(GL_READ_FRAMEBUFFER, GL_COLOR_ATTACHMENT0,
+                                                          texture->GetHandle(), src.level,
+                                                          src.slice);
```
kainino0x: Same here, should we be using src.depth here somehow? (I get if it would be a pain, though, requiring one glReadPixels per layer.)

Kangz:
Thanks for the review @kainino0x, I'll merge this since @Jiawei-Shao answered your concerns.

PR #269, opened 2018-09-03
Vulkan: add macro for vktrace

haoxli:
Hide vktrace layout behide a marco 'DAWN_USE_VKTRACE' and enable it in build if needed.

### DIFF
--- a/src/dawn_native/vulkan/DeviceVk.cpp
+++ b/src/dawn_native/vulkan/DeviceVk.cpp
@@ -421,8 +421,9 @@ namespace dawn_native { namespace vulkan {
             extensions->push_back(extension);
         };
 
-        // vktrace works by instering a layer, but we hide it behind a macro unless we are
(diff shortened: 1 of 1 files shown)


src/dawn_native/vulkan/DeviceVk.cpp
```diff
@@ -421,13 +421,16 @@ namespace dawn_native { namespace vulkan {
             extensions->push_back(extension);
         };
 
-        // vktrace works by instering a layer, so we need to explicitly enable it if it is present.
+        // vktrace works by instering a layer, but we hide it behind a macro unless we are
```
Kangz: It would be nice to explain why, for exmaple: The vktrace layer crashes when used without the vktrace server started, see this vktrace issue <LINK TO THE ISSUE>

Kangz:
Thanks, LGTM and merging without a second review since this is a simple CL.

PR #270, opened 2018-09-03
GN: Use hermetic builds on Linux

Kangz:
This makes the Linux GN builds use the Chromium copy of binutils and
also makes all platforms use the "custom libcxx" which is libc++.

We do this so that the build doesn't depend on anything from the system
on Linux because our bots are still on trusty with very old libstdc++
and linkers.

This change makes all the bots grren! See https://dawn-review.googlesource.com/c/dawn/+/1320

### DIFF
--- a/.gn
+++ b/.gn
@@ -15,9 +15,7 @@
 buildconfig = "//build/config/BUILDCONFIG.gn"
 
 default_args = {
-  use_custom_libcxx = false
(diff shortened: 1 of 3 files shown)


Kangz:
PTAL @kainino0x @SenorBlanco

Kangz:
Thanks!

PR #272, opened 2018-09-03
Descriptorize ComputePipeline

Kangz:
@kainino0x PTAL!

Also you can ignore the VkNonDispatchableHandle commit as it has its own PR.


src/dawn_native/vulkan/DeviceVk.cpp
```diff
@@ -48,7 +48,7 @@ const char kVulkanLibName[] = "libvulkan.so.1";
 #elif DAWN_PLATFORM_WINDOWS
 const char kVulkanLibName[] = "vulkan-1.dll";
 #else
-#    error "Unimplemented Vulkan backend platform"
+const char kVulkanLibName[] = "vulkan-1.dll";
```
kainino0x: `.dll` on non-windows?

Kangz:
Great, thank you!

PR #276, opened 2018-09-06
Use unique_ptr where applicable.

Kangz:
PTAL @kainino0x


src/dawn_native/d3d12/DeviceD3D12.cpp
```diff
@@ -169,12 +169,14 @@ namespace dawn_native { namespace d3d12 {
...
-        delete mResourceAllocator;
-        delete mResourceUploader;
-        delete mFunctions;
+        mCommandAllocatorManager = nullptr;
+        mDescriptorHeapAllocator = nullptr;
+        mMapRequestTracker = nullptr;
+        mResourceAllocator = nullptr;
+        mResourceUploader = nullptr;
+
+        // Delete mFunctions last since it unloads the D3D12 DLLs
```
kainino0x: I believe we could move this comment to the field order, and remove all the explicit deletes here. https://stackoverflow.com/a/2254306

src/dawn_native/Device.cpp
```diff
@@ -50,11 +50,11 @@ namespace dawn_native {
     // DeviceBase
 
     DeviceBase::DeviceBase() {
-        mCaches = new DeviceBase::Caches();
+        mCaches = std::make_unique<DeviceBase::Caches>();
     }
 
     DeviceBase::~DeviceBase() {
-        delete mCaches;
+        mCaches = nullptr;
```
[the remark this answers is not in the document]
Kangz: This one was because I thought this was going to make it free cached objects before the backend device is destroyed. Which is wrong anyway, removed.

src/dawn_native/d3d12/DeviceD3D12.cpp
Kangz: TIL! Done, this makes the code much prettier :D

src/dawn_native/metal/DeviceMTL.mm
```diff
@@ -65,10 +65,7 @@ dawnDevice CreateDevice(id<MTLDevice> metalDevice) {
         [mPendingCommands release];
         mPendingCommands = nil;
 
-        delete mMapTracker;
         mMapTracker = nullptr;
```
Kangz: These one I keep because the device is freed explicitly. When we spend more time on the Metal backend we might want to add an RAII ObjC smart pointer.

src/dawn_native/vulkan/DeviceVk.cpp
```diff
@@ -182,21 +182,13 @@ namespace dawn_native { namespace vulkan {
         }
         mUnusedFences.clear();
 
-        delete mBufferUploader;
         mBufferUploader = nullptr;
```
Kangz: Kept these but added a comment

PR #279, opened 2018-09-10
Make Dawn error macro more explicit and have an "error type"

Kangz:
The error type will help distinguish between validation errors, context
losts and others which should be handled differently.

Take advantage of advantage of this to change DAWN_RETURN_ERROR to
"return DAWN_FOO_ERROR" to have the return be more explicit. Also
removes usage of DAWN_TRY_ASSERT for more explicit checks.


Kangz:
Thanks!

